Issue #11, opened 2013-10-28
drupal_alter($entitytype . '_label') & custom label callbacks


xurizaemon:
In order to make contacts uniquely identifiable, making additional details (eg email, phone) visible in their label can be beneficial.

Exposing [drupal_alter('civicrm_contact_label')](https://api.drupal.org/api/drupal/includes%21module.inc/function/drupal_alter/7) etc permits sites to define their own rules for making unique contacts more visible.

PR #12, opened 2013-10-28
7.x 1.x label callback

xurizaemon:
Refs #11. Turns out drupal_alter() does work, provided you remember to tag `_alter` on the end of your function name.


xurizaemon:
Actually, looking at the PR above, it looks like I threw the customisation into the callback and NOT in drupal_alter. Please don't merge this (how do I cancel a PR?)

PR #33, opened 2014-12-28
Display Suite Integration, display formatters!

jackrabbithanna:
Code to integrate more fully with display suite.  If you enable a display suite layout for a view mode, then CiviCRM properties will respect region, label, and display formatter settings when rendered on the view pages. 
Methodology for adding custom display formatters in place. Image, Date, and "entity linking" formatters now possible.
Also integrated CiviCRM properties on the 'Manage Forms' tab for weight ordering of fields on edit forms..

Added templates for price set, price field, price field values

### DIFF
--- a/civicrm_entity.module
+++ b/civicrm_entity.module
@@ -143,13 +143,13 @@ function civicrm_entity_get_field_type($field_spec) {
     return array('type' => 'serial');
   }
   if (!isset($field_spec['type'])) {
-    return array('type' => 'text');
+    return array('type' => 'text', 'field_type' => 'text');
   }
 
   switch ($field_spec['type']) {
     case CRM_Utils_Type::T_INT:
     case CRM_Utils_Type::T_BOOLEAN:
-      return array('type' => 'integer');
+      return array('type' => 'integer', 'field_type' => 'number_integer');
 
     case CRM_Utils_Type::T_MONEY:
     case CRM_Utils_Type::T_FLOAT:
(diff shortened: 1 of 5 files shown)


jackrabbithanna:
there is one thing that is a bit hacky about this implementation. I needed a hook_form_alter() for the entity manage display form, to set some visibility settings.  I do this through an additional submit handler. Sometimes after enabling a ds layout for a view mode you have to save again for the visibility settings to get in there. I'm not sure what needs to be configured differently to be able to do without this. It does work but I'll continue to research the matter. Without those setting updates, there will be notices on save, and the properties will not show on the view page. 

I'll have an example of some date display formatters soon, so you can see how that's done.

The code adds civicrm properties  to the "Manage Fields" tab fpr weight sorting, not the "Manage Forms" I mentioned in the first comment.

eileenmcnaughton:
Thanks Mark

Where do you think we are at re doing a 2.x release? People are still getting the old code

jackrabbithanna:
If you wanted you could put out an alpha1 to encourage more downloads and get more feedback. 
There are some notices to clean up, but they are not really harmful. View page, edit forms, delete forms, display suite integration, display formatters (just about there).  That's a lot of progress.  There still need to be improvements, like custom fields on edit forms, and some display formatting that will print the value of an option like 'Male' instead of 2 for some fields. But as far as the architecture goes, it is pretty sounds for a 2.x release. From here I plan to clean up the edit forms, and include custom fields, and make some display formatters, as well as fix bugs and do tweaks as they become apparent.

jackrabbithanna:
Have you looked at the new features and tested them at all?  What do you think?  I was hoping from some feed back from you.  What are your thoughts on what a 2.x release should be?

shawnholt:
Sorry guys, I'm plugged in but working hard on another project. Should wrap
ASAP and will try to test this week.  I'm sure civicrm_entity will be the
dominant civi/d7 interface soon!  Great stuff and kudos to both of uiu!

PR #34, opened 2015-01-01
adding handling for 4 new display formatters and set some contact field formatter options on manage display form

jackrabbithanna:
Added handling for:
Entity link, yes/no, true/false, and option value lookup formatters. 
Added formatter options to several fields on civicrm contact manage display form

### DIFF
--- a/civicrm_entity.module
+++ b/civicrm_entity.module
@@ -1864,11 +1864,68 @@ function civicrm_entity_ds_fields_info($entity_type) {
       ),
     );
   }
-    
+  //add specific property formatters
+  _civicrm_entity_add_formatters($entity_type,$fields);
+  
   if (isset($fields[$entity_type])) {
     return array($entity_type => $fields[$entity_type]);
   }
 }
+
+function _civicrm_entity_add_formatters($entity_type,&$fields) {
+   switch($entity_type) {
+   case 'civicrm_contact':
+     _civicrm_entity_contact_formatters($fields);
+     break;
+    
+   }
+}
+
+function _civicrm_entity_contact_formatters(&$fields){
+  $link_fields = array('employer_id_contact');
+  foreach($link_fields as $name) {
+     $fields['civicrm_contact'][$name]['properties']['formatters'] = array (
+       'default' => 'Default',
+       'civicrm_link' => 'Link',
+     );
+  }
+    
+  $option_fields = array(
+    'preferred_communication_method',
+    'prefix_id',
+    'suffix_id',
+    'communication_style_id',
+    'gender_id',
+  );
+  
+  foreach($option_fields as $name) {
+     $fields['civicrm_contact'][$name]['properties']['formatters'] = array (
+        'default' => 'Option ID',
+        'civicrm_option_value' => 'Option Value',
+     );
+  }
+  
+  $yes_no_fields = array(
+    'is_deceased',
+    'do_not_email',
+    'do_not_phone',
+    'do_not_sms',
+    'do_not_trade',
+    'do_not_mail',
+    'is_opt_out',
+    'is_deleted',
+    'contact_is_deleted',
+  );
+  
+  foreach($yes_no_fields as $name) {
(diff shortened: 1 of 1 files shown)


jackrabbithanna:
So you can see that formatters options will need to be set specific on a per entity basis for now, unless someone thinks of a better way

Once the formatter is set the one function function _civicrm_entity_render_fields($field) can handle the generation of the output for all entities based on formatter...I've only gone through most of the contact fields...will chip away at the other entities but I thought I'd look for some feedback

eileenmcnaughton:
I think I noticed the new line at the end of the file is missing - not going to hold things up on that but white space standards for drupal have a blank line at the end of the file for cleaner patching

eileenmcnaughton:
nb - have not done any real review as on holiday :-)

eileenmcnaughton:
actually - getfields should be able to identify 'option_fields' for you too. if a field has a pseudoconstant or an html type of select then  a getoptions call should retrieve the possible values (even if they are a bit 'abnormally constructed') IF type = 16 then that is a boolean

 https://github.com/civicrm/civicrm-core/blob/master/api/v3/examples/Activity/Getfields.php

jackrabbithanna:
well I've got every entity having its own formatter adding function for now, It may be able to be automated some, but I've seen enough variety that eventually you'd want a function per entity. 
Those functions could accommodate other per entity customizations. 
I am using the getoptions api to get the options.  

I don't think that custom fields and other values not returned by the api will be coming soon.  I'm at a point in my thinking about the project that there's a wall now, and the wall is the data that a normal get call returns for an entity in the load() method of the api constructor. This particularly true for the contact entity....maybe most of the rest are not so bad. 

The load method needs to load all the data in the schema. and the save method needs to save it all.

Integrating custom fields will require the automation you are seeking...then after that the per entity specifics....

eileenmcnaughton:
Regarding whether custom fields should be retrieved every load - is it a neck or nothing? Most api will retrieve custom fields if asked nicely - although there are a bunch of Civi inconsistencies. But it also increases the cost of doing the load quite a bit - e.g in views bulk operations.

An alternative is to have another action to load custom data, or, to have a separate entity for each custom data set for an entity - within rules / vbos either of those could possibly work but I'm not sure about other applications

jackrabbithanna:
Yeah I've got some prototype for a secondardy api call that fetches all available custom fields for an entity, then pulls it up...and its slow and there were some issues with the interfaces I have been building.  Once the load() method actually gets all the custom fields, I'll need to redo the display suite integration slightly.  Right now I'm fetching a wrapper with entity_metadata_wrapper for each field and that is horribly in-efficient......Normally you'd just use the entity object that is passed by default....

As your primary strength is in CiviCRM and CiviCRM Api, and my primary strength is Drupal, I was hoping that you could take a swing at including custom fields....Then I'd react to that with the view pages, edit forms, and display suite integration.

eileenmcnaughton:
So, you would argue that 'load' should always load ALL custom fields?

jackrabbithanna:
I would argue that the Drupal entity load should load every custom field associated with the civicrm entity. For one, people are going to want it all right?  From the Drupal entity perspective, it shouldn't matter that the fields is custom or added in addition to stock civicrm fields. We're getting the custom field information on the getproperties call, which is setting up the entity propery info...because custom fields are included in the display suite integration automatically...didn't need to add any additional code for that. The fields are coming through on the Drupal entity edit forms as well, but the data isn't loaded into the form_state.  Setting a value for a custom field, via a Drupal entity edit form save, is working as well...It just that since the load doesn't add custom fields to the CiviCRMEntity object, the data is empty for the view pages, and the field values when editing. If the custom fields were populated, basically, if the results of the CiviCRMEntity object would match what's available with entity_metadata_controller, or get_controller, I could make the display suite render code much more efficient.  The biggest problem right now is this inconsistency between the entity and controller objects.....They should match for the Drupal stuff to all work "as it should".  Match as in having all the same fields and properties and their values

PR #35, opened 2015-01-02
better formatter code organization, some contribution entity fields display formatters

jackrabbithanna:
also a better way to do the link formatters...

and bug fixes

### DIFF
--- a/civicrm_entity.module
+++ b/civicrm_entity.module
@@ -1874,6 +1874,9 @@ function civicrm_entity_ds_fields_info($entity_type) {
 
 function _civicrm_entity_add_formatters($entity_type,&$fields) {
    switch($entity_type) {
+   case 'civicrm_address':
+     _civicrm_entity_address_formatters($fields);
+     break;
    case 'civicrm_contact':
      _civicrm_entity_contact_formatters($fields);
      break;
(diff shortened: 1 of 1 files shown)


eileenmcnaughton:
merging - NB I'm pretty sure you can determine yes_no_fields from the schema metadata / getfields result for a given api

jackrabbithanna:
I could use a code example....The $fields array in those functions don't have any information about meta data.  I couldn't think of a way to automate further whether a fields should have yes/no formatters, link formatters, or option value formatters....Please stop me if there's a better way because I plan on pushing forward with the rest of the entities

jackrabbithanna:
of course what about custom fields...that would need some automation...

I'm trying to do this as quickly as I can for a 2.x release, and holidays are the only time I have off.

jackrabbithanna:
```
$result = civicrm_api3('Contact', 'getfields', array(
'sequential' => 1,
));
```

when i do a getfields api call on a field, i see two bits of information
$result[values][x]['type']
and $result[values][x]['html']['type]

neither of which tells me much about how the field could/should be displayed......

The type is simply whether it is integer, text, or datetime.

As for the HTML type?
Could I say that all 'Select' are option lists? Could all fields with Checkbox html types be yes/no or true/false?

eileenmcnaughton:
Yes, all select fields are option lists and boolean has it's own type (class CRM_Utils_Type::T_Bool). You might be better not using sequential => 1 with getfields - the keys are fieldnames so could be useful.

You can get the options for any field using the getoptions api - at least for custom fields checkboxes could be other than boolean so that would be useful

PR #36, opened 2015-01-02
Rest of entity formatter code, improved link formatter, bug fixes

jackrabbithanna:
Well this will give all the entities some formatter capability.  If we can find a better way to do it great, but its there now. We can maybe do some date formatters all at once, instead of per field. 

link formatter to have entity label as link text now

fixes a couple of template files which were named improperly. 

I think this gives us a solid base for display suite integration.  If we wish to find a more automated way to do this, especially for custom entities, I'm happy to work with you to get it better.

We've got one notice that shows on some entity view pages, especially contact view pages, and I haven't figured how to stop that without stopping the display of those fields which the warning pertain to. 

One big thing to figure out is how to get the custom fields on the entity object in the load method. Then custom field values would appear on view pages, and we could look at testing the edit form widgets for the different custom field types. 

I'm going to wait until you review the code now, and take a deep breath because I feel like this is your project and you may want to evaluate what is going on.


eileenmcnaughton:
I'm going to merge the code, and also add the phone entity based on the ticket. Then I'll try to get my head around how it looks on my local and try to use the metadata a bit more. 

I need to think a bit about the custom fields thing and one of the problems I have is that this is a module that does more than I have a handle on - ie. I use a fairly limited range of functionality from this module but because it is an integration there are a bunch of other uses. Anyway - I guess putting together one or more blogs will be important. I'm going to look at this a bit this week

jackrabbithanna:
I agree it needs some work, and you're proposed approach of using the metadata would be better. I was under a time pressure because I'm going to show this off at the Dallas/Fort Worth CiviCRM meetup and wanted this bit of display suite integration included....

Let me know if you want me to write a blog article, and what to focus on.  

I feel I've added some parts to this project i.e. edit forms and display suite integration that you hadn't originally intended, but we are SO close to having the whole enchilada.....

eileenmcnaughton:
It's not so much that I didn't intend them - I actually don't have a clue what they do :-)

Yes please to a blog - there are probably loads of aspects but perhaps what you are showing at Fort Worth would be a good start?

jackrabbithanna:
Ok I'll start writing up a blog article for civicrm.org.  Maybe kind of a preview. talk about now being able to add drupal fields to an entity, create and manage drupal display suite view modes for Drupal view pages, and talk about support for edit pages being near completed. I've found a couple of issues, one is that the Contribution records won't save...there's no Amount fields coming through the entity meta data controller. But all these things can be overcome, I as feel adding custom fields will be possible as well.

PR #39, opened 2015-02-10
Do not set unused password, use built in notify function

JKingsnorth:
Let's not set a password that isn't used anywhere. We can also use the user module's built in 'notify' function, rather than calling drupal_mail.


JKingsnorth:
@xurizaemon Hi, thanks for the comments.

I've created an issue on Drupal.org here https://www.drupal.org/node/2477089.

I'm not sure how sites are using this functionality, because it is not actually 'working' in the module. The password is not being saved on the user object.

This is for the best since the module is generating an insecure password - this code should definitely be removed. If it was in use anywhere I would have opened a security issue instead.

Edit: At the very least, if this functionality is still needed, it should use [user_password()](https://api.drupal.org/api/drupal/modules!user!user.module/function/user_password/)

PR #41, opened 2015-04-07
fix bug loading multiple entities

jackrabbithanna:


eileenmcnaughton:
merged - as an aside - did you spot https://github.com/eileenmcnaughton/civicrm_entity/pull/40

jackrabbithanna:
I saw that you had patch, but I haven't delved into it yet....I would need to apply this patch to 4.6? or will it work for earlier versions?
looks like a big patch.....

eileenmcnaughton:
well - the api gets consistently better at loading values - so it's more like some entities would work better on 4.6 than earlier civi - possibly

jackrabbithanna:
or can I just clone the master branch of the civicrm project, and will that have the right code for testing the custom fields functionality in civicrm_entity?

eileenmcnaughton:
I'd just try the patched civicrm_entity against whatever civi you have & see how good it is - the syntax in use has been in use for a long time - it's just that a load of bugs have been ironed out over a long time too

jackrabbithanna:
I've been testing on 4.6 in preparation for my CiviCon presentation. No dice as is with custom field values getting loaded on the edit forms, display suite property formatters, or view pages...
Making a civicrm_entity a big part of that presentation, as well as 2 other modules we've released recently: https://denver2015.civicrm.org/sessions/civicrmdrupal-integration-showcase
will you be at CiviCon?

jackrabbithanna:
oh, you haven't merged that patch into 2.x branch yet? I'll give it a whirl and will let you know

jackrabbithanna:
ok well that patch is a good start, the data is coming through! So the entity object will get populated with the custom field values!

The type information is still not coming in, so the widget for the edit form and the property formatters for the display suite integration still don't get set, but at least the data is there...may want to go ahead and merge that into 2.x and let the community beat on it a little bit

jackrabbithanna:
well now wait a sec, a custom field of type date is working on the edit forms....I'll have to dig a little deeper here, this may just need some tweaking and we can have it all

eileenmcnaughton:
It might just be a case of mapping the type info across. Let me know when you want me to merge

jackrabbithanna:
as far as your patch goes, bringing in the custom field data will make a lot of people happy, I'm seeing tokens being generated automatically, which should make it accessible to rules...the data is coming through the tokens in display suite fields :)

I think I'll be able to make this work.....its going to take me several hours of work and more of testing but should be able to happen now for the edit pages and display suite integration...look for a patch in a few days....

As far as working with 4.6 goes its looking real good, patch it in! Everything may not work with every version of CiviCRM but oh well.

eileenmcnaughton:
done.

Keep an eye out for the fact custom fields declare 'data_type' & core fields declare 'type' - might need to ensure data_type is interpretted in the schema hook

jackrabbithanna:
are the values for the type the same for both core fields and custom fields?

eileenmcnaughton:
no - custom_fields have a field in the civicrm_custom_field table. You could make a very good case that the api should add a field 'type' for each custom field with a translation & if it does it probably doesn't do that on all civicrm versions

jackrabbithanna:
so the constants used  in the civicrm_entity_get_field_type() function such as CRM_Utils_Type::T_TEXT won't work for custom fields?

eileenmcnaughton:
here is an example of what getfields looks like for a custom field https://github.com/civicrm/civicrm-core/blob/master/api/v3/examples/Contact/GetFieldsOptions.php#L729

eileenmcnaughton:
So , at least in 4.6, type IS declared...

jackrabbithanna:
yeah, so the ['type'] index value for the custom_1 is 2, same as in some core fields, so that's varchar I guess....that is what the civicrm_entity_get_field_type() is looking at to define the schema type. But I don't think this really matters, since drupal isn't doing any of the actual CRUD, the CiviCRM api is. 
So actually I don't think the schema code will need any alteration.....the schema should only matter insofar as form validation is concerned, which is a whole other thing. I'll need to test to see what happens if the user enter's a letter 'a' instead of a 1 for a field that is an int even though the widget would be the same.  the api would puke, and no harm would be done, the thing is to give proper feedback to the user as to what to fix....
There's a lot to think about, and it'll be an incremental process....but I'm making some custom field widgets work already so its going to work out in the end. 

I wish I could spend a solid week or two on this, but all this pesky client work keeps me too busy :)
As it is though I feel like the dam has burst and I have a lot now that I can do....Thanks for making the civicrm custom field data come into the entity object, that was really holding me back. 

the widgets and property formatters need to know the ['data_type'] or the ['html']['type'], I can handle that.

jackrabbithanna:
there could be field Pre Help and field post help text to do as well......is there any api thing for that?

eileenmcnaughton:
Cool - btw - yes to CiviCon.

I think to get pre & post field help we'd need to do a get call on the custom field

jackrabbithanna:
well, I guess we can cross the pre and post field help text river after we get the widgets in place...

Is there anything you want to say about CiviCRM Entity in the presentation? I can share my slides with you to review if you like. shoot me an email at mark@skvare.com with the email address you'd like to receive the slides at...

jackrabbithanna:
Is there any thought at all about getting CiviCRM Entity into the core drupal module? The more I think about it, the more I think that it should be there...

eileenmcnaughton:
Good conversation for CiviCon - my focus had been on getting it in for Drupal 8 - although that is a little stalled at the moment

jackrabbithanna:
This project could really use a PM tool.....
Here's some notes:
Done:
All the basic custom fields types are handled fairly easily on the edit forms, text,select, multi-select, radios, checkboxes,  could post a patch for those...

CiviCRM Entity TODO:

Edit forms:
1)autocomplete select widget
2)handle custom fields that accept multiple values (on a custom field set level) wtf to do there?
Should custom field sets be fieldgroups and have custom fields organized in those...
but drupal doesn't do multiple field values by field group...so custom field sets that have multiple values this may need to be a custom widget

3)field type Yes and No shouldn't have the "no value" option
When should other select fields not have this option?

4)file
5)link
6)contact reference custom field types.....may need a custom widget for this too....borrow from entity reference? or use the core drupal contact reference field?

Display Suite Integration

Going to take a real hard look at how I'm doing display suite integration "formatters" features now, and try to see if there is not a more condensed and general method to be had, because what is there won't work for custom fields

PR #42, opened 2015-04-11
fix so that civicrm_contribution source displays properly

jackrabbithanna:
For the Drupal entity labels, you have to set a property that is on the Drupal entity object. For some reason, the source column is coming through to the Drupal entity object as 'contribution_source'.


jackrabbithanna:
Also check out commit a57ab5d

To get the total_amount and other decimal/money fields I needed to add these two lines.  It worked for including fields of these types in the entity_metadata_wrapper(). 
Was there a reason not to include these before?  Is 'decimal' the right value?

PR #51, opened 2015-07-30
civi case support

jackrabbithanna:
Also I fixed a bug that was occuring when I was trying to view the label of entities referenced using entity reference fields, in Views. 

check out the else clause in function_civicrm_entity_access line 67

I call the other entity access callback to handle the cases better than just defaulting to "administer CiviCRM"

### DIFF
--- a/civicrm_entity.default_form.inc
+++ b/civicrm_entity.default_form.inc
@@ -1,0 +1,219 @@
+<?php
+
+function civicrm_entity_form($form, &$form_state, $entity_id, $op, $entity_type) {
+
+  if ($op == 'edit' || $op == 'delete') {
+    $entity = entity_load($entity_type, array($entity_id));
+    $entity = $entity[$entity_id];
+    $form_state['entity'] = $entity;
+    $form_state['entity_type'] = $entity_type;
+  }
+  if ($op == 'create') {
+    $entity = new CiviCRMEntity(array(), $entity_type);
+    $entity->is_new = TRUE;
+    $entity->type = $entity_type;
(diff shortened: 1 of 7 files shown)


jackrabbithanna:
Wondering if you wanted CiviCase support, and this bugfix to make it into the main branch....

I'm working on some new stuff as well.....

jackrabbithanna:
Added support for CustomField and CustomGroup entities
Fixed display suite integration so that layout css is loaded when entity is viewed
reorganized code, put the form stuff in its own file
put the display suite stuff in its own file
fixed a couple of notices

eileenmcnaughton:
Hey Mark, sorry I've only been partly following due to workload. Is this ready for merge now? Is this the point where we should change version?

jackrabbithanna:
I think everything that is this pull request could stay in the 2.0 branch....I'm working on a different approach to custom fields, but i"m doing that in a separate module for now, and it may require a patch to the Drupal Entity API module....I'm experimenting with some different ideas...These commits fixes a couple of bugs, does a bit of code reorganization, and exposes CustomField and CustomGroup entities to Drupal.  I need these exposed for the future work that I'm doing....which may include adding custom groups and fields from the Drupal Manage Fields page.....

jackrabbithanna:
FYI: https://github.com/jackrabbithanna/civicrm_entity_custom_values
Requires the modified Entity API module ATM....provided in the repo

eileenmcnaughton:
Ok - do you want me to merge this now - or hold off for a bit longer?

PR #52, opened 2015-10-30
Better labels.

xurizaemon:
Currently we see "CiviCRM Entity_tag" but it's nicer to see "CiviCRM Entity Tag".


xurizaemon:
Note also your comments on f85742c on my branch which is long forgotten, and not required for this, but is a label-ish thing. That relates to the entity label ("Chris Burgess" rather than the label for the entity type ("CiviCRM Contact").

https://github.com/xurizaemon/civicrm_entity/commit/f85742c76e9daffa1baa28b2674900eb844086b0

PR #53, opened 2015-11-30
VBO fixes

aydun:
Fixes for working with VBO:
Make civicrm_entity_access() use more extensive checks in civicrm_entity_op_access() - relates to https://www.drupal.org/node/2505111
civicrm_entity_access is called by VBO so this lets VBO update more without requiring 'administer CiviCRM' perms.
Also remove 'administer CiviCRM' requirements relating to contributions

'Administer CiviCRM' enables any action - not sure that's the right behaviour but changing it will probably break existing setups.

### DIFF
--- a/civicrm_entity.module
+++ b/civicrm_entity.module
@@ -54,18 +54,13 @@ function civicrm_entity_permission() {
  * @return bool
  */
 function civicrm_entity_access($op, $entity, $account, $entity_type) {
-  if ($op == 'view' && $entity_type == 'civicrm_event') {
-    return user_access('view event info');
+  // expected $op values for civicrm_entity_access when called from VBO: view, update, create, delete
+  // expected $op values for civicrm_entity_op_access: view, edit, create, delete
+  if ($op = "update") {
+    $op = "edit";
   }
-  elseif ($op == 'view' && $entity_type == 'civicrm_participant') {
-    return user_access('view event participants');
-  }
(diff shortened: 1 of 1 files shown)


eileenmcnaughton:
This looks Ok to me - @xurizaemon see anything you have doubts about - or merge-ahoy!

jackrabbithanna:
Hey guys, I'm making a huge commit soon, and this issue will be come moot.
I've created a hook hook_civicrm_entity_supported_info(&$civicrm_entity_info) 
That will allow CiviCRM Extensions, or Drupal modules to implement it. They will be able to then add new entities, or alter metadata to vary the permissions necessary for all ops, the display suite integration, the theme template. Its very cool. 

This way we can have a standard baseline set of permissions, but module developers can alter that to use whatever permissions they want for the view, update, edit, create, delete operations, for each entity. It falls back on 'administer CiviCRM' if no permissions are specifically listed for any op of an entity. It basically does what this merge request is for, and more. 

@aydun there will be a civicrm_entity.api.php that clearly documents the options for this new civicrm entity metadata info alter hook.

aydun:
@jackrabbithanna sounds great.  When its out I'll re-implement my use case.

jackrabbithanna:
The 2.x branch here is updated, as well as a new 3.x branch.  From here on 2.x will be mainly bug fixes and small feature improvements, but nothing too radical.  3.x will have the new goodies

PR #54, opened 2016-02-15
Identify fields, filters, etc. coming from civicrm_entity moduel to avoid confusion with fields provided by core's implementation of hook_views_data().

torrance:
civicm_entity provides duplicate fields and filters for contacts, contributions, etc. etc. and they don't work with core's fields. I'm unsure for the reason for providing a duplicate implementation of these fields.

For now, this patch helps identify fields, filters, etc. coming from civicrm_entity moduel to avoid confusion with fields provided by core's implementation of hook_views_data().

### DIFF
--- a/civicrm_entity_default_views_controller.inc
+++ b/civicrm_entity_default_views_controller.inc
@@ -76,7 +76,7 @@ class CiviCRMEntityDefaultViewsController extends EntityDefaultViewsController {
 
     $description = array(
       'title' => $property_info['label'],
-      'help' => isset($property_info['description']) ? $property_info['description'] : NULL,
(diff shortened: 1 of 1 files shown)


torrance:
To clarify 'they don't work with core's fields':

It's buggy if you start building a view based on a core field and then use a relationship to link in a field but accidentally use a field provided by civicrm_entity. The issue is that there are two data descriptions for the same set of tables but they're not necessarily related in the same way and this can generate SQL errors.

The other issue is just one of usability: ie. why do I have two identical fields on offer in views (with the exception that civicrm_entity has an error message in place of the help text).

PR #60, opened 2016-03-13
adding new administer CiviCRM Entity permission, changing entity mana…

jackrabbithanna:
…gement paths to outside of content types


jackrabbithanna:
New root path: admin/structure/civicrm-entity

example entity management paths:
Fields:
admin/structure/civicrm-entity/civicrm_contact
Display settings
admin/structure/civicrm-entity/civicrm_contact/display/full

PR #69, opened 2016-09-15
Event registration form and processing for CiviCRM Entity Price Set Field

jackrabbithanna:
also beginning to expose all the financial entities for Drupal views based financial reporting


jackrabbithanna:
price set field now offers event registration on the display of the event entity
Working with line items via the transact Contribution api action. 
Creates contribution, participants, participant_payments and line_items
Needs testing with price sets with multiple price fields

PR #75, opened 2016-09-23
Issue #2476777: Coding standards for 7.x-2.x, misc

xurizaemon:
A place to stash code cleanup that isn't related to any other specific ticket.

### DIFF
--- a/civicrm_entity.install
+++ b/civicrm_entity.install
@@ -8,35 +8,34 @@
 /**
  *  Implements hook_uninstall()
  *
- */ 
+ */
 function civicrm_entity_uninstall() {
   // We have to do each entity individually because we don't have the functions in the module file available after module disable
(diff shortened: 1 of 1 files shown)


jackrabbithanna:
Good to bring to my attention, don't mind the change, but the number of entities is now 40....we need to add the rest to the hook_uninstall() implementation
see civicrm_entity_supported_entities_info()

jackrabbithanna:
Hmm, actually since other modules or extensions can also expose entities we need a more automated way to do this.
Gonna need to call civicrm_entity_get_supported_entity_info() and go through each entity and delete the fields

xurizaemon:
All good. Probably something like -
- Get our list of things (civicrm_entity_supported_entities_info() or whatever)
- Provide a hook for other modules to add theirs
- DELETE ALL THE THINGS

PR #84, opened 2016-10-10
issue https://www.drupal.org/node/2816109

Vgratioulet:
When displaying a CiviCRM custom field in a Display Suite view mode, in display or form,
and trying to export via features,
it can export the fields only in another site where the custom fields have the same ids

if you have two sites that are creating CiviCRM custom fields, they can't export view modes to one another.

Here are whanges to handle the custom fields by names instead of ids
I was able to test it in display mode and form
but not in views where I can't get to display any custom field managed by CiviCRM Entity, with any version of civicrm-entity

### DIFF
--- a/civicrm_entity.default_form.inc
+++ b/civicrm_entity.default_form.inc
@@ -398,15 +398,14 @@ function civicrm_entity_form_submit(&$form, &$form_state) {
   $entity_type = $form['#entity_type'];
   $entity = $form['#entity'];
   civicrm_initialize();
-  $civicrm_entity_fields = civicrm_api(substr_replace($entity_type, '', 0, 8), 'getfields', array(
-    'version' => 3,
-    'action' => 'create',
-  ));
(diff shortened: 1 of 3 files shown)


jackrabbithanna:
Can you reroll your patch?  Several other updates have come through and caused a merge conflict

Vgratioulet:
Here it is

jackrabbithanna:
This code will not work with CiviCRM 4.6, only 4.7 .... which is fine I suppose.  I'm a little concerned that this PR will break people's Rules upon upgrade, the entity metadata property name for the custom fields will be different...but it would be nice to be able to export custom fields....hmmm thinking

Vgratioulet:
You're right, if it breaks some existing things that's an issue
I can try to write an update function for that, 
I haven't used Rules yet, can you point me in the right direction to have a test rule with a custom field, so I can see if i'm able to write that function ?

PR #102, opened 2017-02-22
Creating in the same form contact and linked entities

Vgratioulet:


jackrabbithanna:
just a quick scan, and I wonder if the updates in the civicrm_entity.default_forms.inc needs to check if the field type of the inline entity form widget is of type civicrm_entity_reference, because this may not be necessary for regular entityreference fields...

jackrabbithanna:
I'd also like to confirm if the second $wrapper->save() on 478 is absolutely necessary.  In a perfect world we'd only save each entity once, it may not be possible, but if we can that should be the way it is. I don't think its a deal breaker though

Vgratioulet:
checking the field type of the inline entity form widget in the civicrm_entity.default_forms.inc is a good question. Regular entityreference fields could be also linked to the main entity. I think that if they're not, the test on line 456 ' if ($host_entity_type == $entity_type) {' will exclude them. 
I agree it would have to be tested, linked and not linked. 

about the second $wrapper->save() on line 478, if you consider the first is on line 463, they're in a if/else condition, so only one of them is run. Or did I overlooked another one ?

diskbuster:
i really need a solution for that issue.

Iam asking me what will happen to the form when using the CiviCRM Entity Reference Field with the CiviCRM Entity Reference Default Widget. Will there be the input fields as chosen in the entity form display?
at the moment the widget gives me some errors (latest dev.) 
Notice: Undefined index: op in civicrm_entity_reference_field_field_widget_form_alter() (Zeile 661 von /sites/all/modules/civicrm_entity/modules/civicrm_entity_reference_field/civicrm_entity_reference_field.module).

If you need someone for testing, i be there.

jackrabbithanna:
@jokas , this bit is next on my list when I have a 4-6 hour block of free time, hopefully this weekend...

Vgratioulet:
Can you give more details about your use case please ? : which entities on which form, what parameters for the inline entities ?
It works for me with inline phone, email and instant_messaging on the contact form, each one with 'Inline entity form - Single value', 'Allow users to add new entities.' is checked and number of values 'unlimited'

diskbuster:
@Vgratioulet I came here over the following thread, where i made a longer description.
http://civicrm.stackexchange.com/questions/17823/adding-referenced-civicrm-entity-contact-in-node-with-email-address-and-phone-d

I made a nearly clean install and checked additionally your settings with no success.

If you like, I can give you access to the system, so you can see whats up.

PR #124, opened 2017-11-13
Allowing admins to disable entity integration by API entity

jackrabbithanna:
Pretty big commit here, adding a admin config page for CiviCRM Entity, allowing admins to disable entity integrations....All submodules packaged with CiviCRM Entity will automatically enable entity types as they are required. 

admin/structure/civicrm-entity/settings

### DIFF
--- a/civicrm_entity.admin_form.inc
+++ b/civicrm_entity.admin_form.inc
@@ -1,0 +1,65 @@
+<?php
+
+/**
+ * @file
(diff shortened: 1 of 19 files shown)


jackrabbithanna:
Going to ship this to Drupal.org as the next beta....

Then after that, we're going to release 2.0 official/stable by the first of the year.

PR #126, opened 2017-11-22
Adding First.Middle.Last to username format options

magnolia61:
In our organisation we had a need to include the middle name in the username format for the creation of a new linked drupal user account.
Since i'ts only a possible to select from a drop-down or use 1 of the data select fields we could not construct this via the ui. Hence we added it to the username format options.


jackrabbithanna:
yeah this looks fine, PR accepted, merging #126

jackrabbithanna:
This patch has been pushed to the -dev version on the project page on drupal.org/project/civicrm_entity

PR #127, opened 2017-12-06
Conditional if middle name is empty

magnolia61:
In my previous PR (https://github.com/eileenmcnaughton/civicrm_entity/pull/126) added a new option to be able to create to new linked drupal user account with the syntax firstname.middlename.lastname I forgot to add a mechanism in case the middle name is empty. So we ended up in usernames like firstname..lastname (note the 2 dots). Therefor this new PR.


jackrabbithanna:
Ok thanks for the fix!

PR #129, opened 2017-12-09
Issue #2806011. Avoid calling drupal_goto() directly from rules actions.

xurizaemon:
See https://www.drupal.org/project/civicrm_entity/issues/2806011

### DIFF
--- a/modules/civicrm_entity_actions/civicrm_entity_actions_action_schedule.inc
+++ b/modules/civicrm_entity_actions/civicrm_entity_actions_action_schedule.inc
@@ -41,26 +41,26 @@ function civicrm_entity_actions_action_schedule_action_info() {
 
 function civicrm_entity_actions_action_schedule_add_action($entity, $context = array()) {
   $base_url = str_replace('_', '-', $context['entity_type']); // civicrm-action_schedule
-  rules_action_drupal_goto($base_url . '/add');
+  civicrm_entity_drupal_goto($base_url . '/add');
 }
 
 function civicrm_entity_actions_action_schedule_view_action($entity, $context = array()) {
   $info = entity_get_info($context['entity_type']);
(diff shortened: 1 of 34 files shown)


modules/civicrm_entity_actions/civicrm_entity_actions_action_schedule.inc
```diff
@@ -41,26 +41,26 @@ function civicrm_entity_actions_action_schedule_action_info() {
 
 function civicrm_entity_actions_action_schedule_add_action($entity, $context = array()) {
   $base_url = str_replace('_', '-', $context['entity_type']); // civicrm-action_schedule
-  drupal_goto($base_url . '/add');
+  rules_action_drupal_goto($base_url . '/add');
```
jackrabbithanna: If you use this function then this module will need to add a dependency to Rules in the .info file

jackrabbithanna: alternatively you could add a if (module_exists('rules')) { new function, else drupal_goto

jackrabbithanna:
@xurizaemon I'm going to be cutting 7.x-2.0 stable next week. If you have time, and want to finish this off, we can get it in the 2.0 release

xurizaemon:
Sorry I'd missed yr earlier comments. Second approach seems sensible.

IDK if this is really a problem for anyone in the real world, so I'm OK with it not being merged!

I had five minutes to implement your suggestion but I don't have time / available test sites to test this with right now.

jackrabbithanna:
Your solution seems great for this in your second commit.  I'll test this and it'll go into 2.0. Thanks alot :)

xurizaemon:
no-worries-alot

jackrabbithanna:
Merging #129

PR #130, opened 2017-12-20
[WIP] 8.x 3.x MVP

mglaman:
This adds an MVP which uses the Event entity. It has 

* basic field mapping to Drupal 8
* Supports loading
* Supports edit form
* Supports delete
* Basic PHPUnit tests

Need to 

* test Create form
* Add field UI for customizing the field widgets on form display
* Improve forms in general
* Tackle noted `@todos`


dsnopek:
Awesome, thanks! I'll do some testing and review soon :-)

xurizaemon:
Cool! I wonder if, since this wraps the CiviCRM APIv3, you'd want to version the classes here (CiviCrmApiV3Interface etc) to futureproof.

:thumbsup:

jackrabbithanna:
@xurizaemon good question....we'd want to use apiv4 when it is finished, or be able to swap them in or out.

I'm going to go ahead and merge this so we have something we can all fork and participate. This is awesome

Thank you!

jackrabbithanna:
The first thing I'm going to be exploring, is the possibility of using this as a template, and provide mechanism to define a number entity types by default, and let developers expose additional entities by hook or plugin...

This can be done in D8 right?  In D7 we had hook_entity_info() and we could get a big list of supported entity types, and use one set of entity api objects that worked for making the integration for all of them work without having to have a set of files for each entity.

dsnopek:
@jackrabbithanna I don't think @mglaman really intended for this to get merged yet! He marked it as WIP and had some stuff he still wanted to fix...

mglaman:
Injected services was an incoming item. As were a bunch of the follow-ups. but I guess I can just start pushing PRs for each of those.

dsnopek:
@jackrabbithanna To answer this question:

>This can be done in D8 right? In D7 we had hook_entity_info() and we could get a big list of supported entity types, and use one set of entity api objects that worked for making the integration for all of them work without having to have a set of files for each entity.

I think this could be done with a deriver to generate the entity type plugins. We'd just need a generic class and a way to build up the plugin definition based on the CiviCRM API data. Here's some generic docs on plugin derivatives:

https://www.drupal.org/docs/8/api/plugin-api/plugin-derivatives

That said, I haven't actually tried doing derivatives with entity types - I've mainly used it with blocks and layouts.

dsnopek:
Oh, hey! Just started reviewing and it looks like @mglaman's code already accounts for this? :-) He's using a `hook_entity_type_build()` to dynamically declare entity types based on a list of supported types. Although, I'm not sure then why the `Event` class has a `@ContentEntityType` annotation? Shouldn't that be unnecessary with the `hook_entity_type_build()`?

src/CiviEntityStorage.php
```diff
@@ -0,0 +1,181 @@
+<?php
+// In construct make sure to invoke initialize
```
dsnopek, line 2: Stray comment?

src/CiviCrmApi.php
```diff
@@ -0,0 +1,54 @@
...
+  /**
+   * Constructs a new CiviCrmApi object.
+   *
+   * @param \Drupal\civicrm\Civicrm $civicrm
+   *   The CiviCRM service.
+   */
+  public function __construct(Civicrm $civicrm) {
+    // Ensure CiviCRM is loaded and our function is available.
+    if (!function_exists('civicrm_api3')) {
+      $civicrm->initialize();
```
dsnopek, line 18: Don't do this in the service constructor! It'd be better to call `$civicrm->initialize()` when we're about to do work with CiviCRM. Otherwise, we're initializing CiviCRM any time this service is constructed, and it might be constructed just to resolve a dependency, but when we're not actually going to do anything with CiviCRM.

For example, querying info about the entity type could lead to the entity type plugin being loaded and then initializing CiviCRM, but maybe it was just some generic Drupal code trying to get some generic entity type info and not really load anything? Initializing CiviCRM is on par with bootstrapping Drupal and shouldn't be done unless we really need to!

I'd suggest putting this in a helper method and then call that at the top of each of the methods that do stuff like `get()`, `delete()`, etc, so it's only initialized just before actually doing something.

src/CiviEntityStorage.php
```diff
@@ -0,0 +1,181 @@
...
+    if ($entity->isNew()) {
+      // @todo decide on special handling?
+      $return = SAVED_NEW;
+    }
+    else {
+      $return = SAVED_UPDATED;
+    }
+
+    $params = $entity->toArray();
+    $params = array_map(function ($value) {
```
dsnopek, line 67: It'd be better if this code was more explicit, ie. looked at field values and decomposed them based on their expected structure. We should be able to know how a field will be represented in advance (based on how it was defined), so if it's going to be `[ 'und' => [ 0 => [ 'value' ] ] ]`, let's explicitly take that apart.

src/Entity/Events.php
```diff
@@ -0,0 +1,26 @@
...
+namespace Drupal\civicrm_entity\Entity;
+
+use Drupal\Core\Annotation\PluralTranslation;
+use Drupal\Core\Annotation\Translation;
+use Drupal\Core\Entity\Annotation\ContentEntityType;
+
+/**
+ * Defines the comment entity class.
+ *
+ * @ContentEntityType(
```
dsnopek, line 12: Do we need this? Shouldn't the `hook_entity_type_build()` replace the need for an annotation?

src/Entity/Query/CiviCRM/QueryFactory.php
```diff
@@ -0,0 +1,53 @@
...
+class QueryFactory implements QueryFactoryInterface {
+
+  /**
+   * The namespace of this class, the parent class etc.
+   *
+   * @var array
+   */
+  protected $namespaces;
+
+  protected $civicrmApi;
```
dsnopek, line 19: Needs a doc comment!

src/Entity/Query/CiviCRM/QueryFactory.php
```diff
@@ -0,0 +1,53 @@
+<?php
+
+namespace Drupal\civicrm_entity\Entity\Query\CiviCRM;
+
+use Drupal\Core\Entity\EntityTypeInterface;
+use Drupal\Core\Entity\Query\QueryBase;
+use Drupal\Core\Entity\Query\QueryException;
+use Drupal\Core\Entity\Query\QueryFactoryInterface;
+
+class QueryFactory implements QueryFactoryInterface {
```
dsnopek, line 10: Oh, and this one too :-)

src/Entity/Events.php
```diff
@@ -0,0 +1,26 @@
+<?php
+
+namespace Drupal\civicrm_entity\Entity;
+
+use Drupal\Core\Annotation\PluralTranslation;
+use Drupal\Core\Annotation\Translation;
+use Drupal\Core\Entity\Annotation\ContentEntityType;
+
+/**
+ * Defines the comment entity class.
```
dsnopek, line 10: This isn't a comment entity :-)

src/SupportedEntities.php
```diff
@@ -0,0 +1,1459 @@
...
+
+namespace Drupal\civicrm_entity;
+
+/**
+ * Defines supported entities.
+ *
+ * See `$civicrm_entity_info['civicrm_event']` for implemented usage. This might
+ * not be required. Perhaps we just implement entity class definitions using
+ * this information. Providing entity type derivatives seems hard, and possibly
+ * overhead when we just can provide the classes directly.
```
dsnopek, line 11: Hm, ok, I guess we do need the annotation or derivatives? It'd be best not to have to make a class with an annotation because we're talking about A LOT of classes (dozens?). If that needs derivatives, I think that's what we should do.

dsnopek (commented):
Posted some review! I think it would have been easier if this wasn't merged already, because that would have allowed a review and update and repeat cycle, but c'est la vie :-)

jackrabbithanna:
Well its merged now, and he can still fix it

jackrabbithanna:
Sorry about merging that too quickly guys....I got excited...The branch was completely empty, I didn't see the harm....

src/SupportedEntities.php
```diff
@@ -0,0 +1,1459 @@
...
+        'edit' => [],
+        'update' => [],
+        'create' => [],
+        'delete' => ['delete activities'],
+      ],
+      'theme' => [
+        'template' => 'civicrm-activity',
+        'path' => drupal_get_path('module', 'civicrm_entity') . '/templates',
+      ],
+      'display suite' => [
```
jackrabbithanna, line 38: I'm wondering if the display suite section of the supported entities metadata will be necessary now.  I wonder if we aim to support the new Layout experimental modules in 8.4 and 8.5.  Not an expert on the nuances of display suite vs. the new Layout stuff going into Core soon. @mglaman your thoughts on this?  Do we need Display Suite anymore to accomplish what we did with the D7 version.....I used Display suite integration in the D7 version so that entity properties could have display formatters. Is this necessary now?

jackrabbithanna:
So very cool, you guys are my heros...
I see data at [site_url]/civicrm-event/1
edit form:
[site_url/admin/structure/civicrm-entity/civicrm-event/1/edit

Can we make the edit form at:
http://mydropdemo.me/civicrm-event/1/edit ? 

Is there currently a path to rearrange the fields...like a manage fields page for events, or is that still forthcoming?

mglaman:
That is coming.

src/SupportedEntities.php
dsnopek: IMO, it's hard to say what will win out at this point. Layout Discovery (new core module) is for sure a thing: it's used by Display Suite, Panels and Field Layout (the core DS-like thing). Why not support them all? ;-)

I don't know DS or Field Layout that well, so I'm not sure what's necessary to integrate with them. But if we do the Entity/Field API right, then Panels/Panelizer should mostly just work (a plugin is necessary to say the entity type supports Panelizer - I can help with that bit. :-))

jackrabbithanna: Definitely want to support the Core Layout Discovery, but of course been loving,knowing.loving  Display Suite in D7,...I know you'lls interest in this is maybe the REST API / Services side of things, but I've spent a a ton of time, and have a lot of clients buying into front side things, like I've replicated the civicrm event registration process in Drupal 7 FAPI , and I want to do the contribution form too for the D8 version. This way us Drupal folk can have Drupal FAPI forms we can customize our way :) The way it sounds with Layout discovery and all those modules using it, we can have several cakes and eat them too. I'm excited to be able to start porting my D7 FAPI custom field types to D8

jackrabbithanna: I got an eye toward some some Commerce integration too...

jackrabbithanna:
@mglaman  :) thank you for working on this

PR #133, opened 2018-01-04
Prefer matching against Individuals, but if not, match as previously.

aydun:
If an Individual and Organisation have the same email address, the match
fails as it is not unique. Drupal logins should normally be linked to
Individuals, so look for those first.


jackrabbithanna:
Thanks for contributing, I'll review the patch. I wonder if we should put an option in Rule action config...

jackrabbithanna:
I'm going to accept this, I think generally people have user accounts, not organizations...Merging #133

PR #134, opened 2018-01-05
Created Drupal username does not get trimmed well

magnolia61:
When a user enters a first,middle or lastname with an extra space in front or at the end the it will be replaces by a _ in the final username. Like for instance firstname_.lastname

I solved this with this PR
`$params['name'] = _civicrm_entity_clean_login_name(trim($contact['first_name']) . "." . trim($contact['last_name']));`
I'm not sure if my solution is the most efficient way of doing this.
Another option maybe would be to use the clean function on all subparts like this:
`$params['name'] = _civicrm_entity_clean_login_name($contact['first_name'] . "." . _civicrm_entity_clean_login_name($contact['last_name']));`


jackrabbithanna:
I think this is fine, thanks for contributing! Merging #134

PR #139, opened 2018-01-23
Fixups

mglaman:
More clean up


civicrm_entity.routing.yml
```diff
@@ -5,3 +5,11 @@ civicrm_entity.admin:
...
   requirements:
     _permission: 'access administration pages'
+
+civicrm_entity.settings:
+  path: '/admin/structure/civicrm-entity/settings'
+  defaults:
+    _form: '\Drupal\civicrm_entity\Form\CivicrmEntitySettings'
+    _title: 'Settings'
+  requirements:
+    _permission: 'access administration pages'
```
dsnopek: This should have a more specific permission - probably a custom settings permission from this module. We usually give site managers "access administration pages" just to get `/admin` and then more specific permissions for other stuff

civicrm_entity.routing.yml
```diff
@@ -5,3 +5,11 @@ civicrm_entity.admin:
     _title: 'CiviCRM Entity'
   requirements:
     _permission: 'access administration pages'
```
dsnopek: This permission too, but... Is this route still needed?

dsnopek (commented):
Perfect!

EDIT: Er, I was trying to comment on commit https://github.com/eileenmcnaughton/civicrm_entity/pull/139/commits/e400fcbc194460265c7269ed0a5949be4e951a57 - I'm too used to using GitLab which lets you comment on a commit ;-)

src/CiviEntityStorage.php
```diff
@@ -63,21 +64,33 @@ protected function doSave($id, EntityInterface $entity) {
...
         }
-        return $value;
+        $list[$delta] = $item;
+      }
+      // Remove the wrapping array if the field is single-valued.
+      if ($storage_definition->getCardinality() === 1) {
+        $list = reset($list);
+      }
+      if (!empty($list)) {
+        $params[$field_name] = $list;
```
dsnopek, line 90: All these changes in `doSave()` look great too! It's clear what this code is doing and why :-) :+1:

dsnopek:
Overall, this looks awesome, thanks! This is super great clean-up! The only actionable comment I have is the one about permissions on the admin routes above

civicrm_entity.routing.yml
jackrabbithanna: in the D7 version we have 'administer CiviCRM Entity'

mglaman:
Now `administer civicrm entity` is used for routes and entity admin.

tests/modules/civicrm_entity_test/migrations/civicrm_events.yml
```diff
@@ -0,0 +1,28 @@
...
+migration_tags:
+  - 'CiviCRM Entity'
+source:
+  plugin: embedded_data
+  ids:
+    title:
+      type: string
+  data_rows:
+    - title: 'DrupalCon Nashville'
+    - summary: 'DrupalCon Nashville 2018! Get your Drupal on'
```
dsnopek: The extra '-' in front of 'summary' here seems not right. If I'm reading this correctly, this would make two data rows like this:

```
[
  [
    'title' => 'DrupalCon Nashville',
  ],
  [
    'summary' => 'DrupalCon Nashvilee 2018! Get your Drupal on',
    'description' => '...',
    // etc
  ]
]
```

mglaman:
Oops, you're right!

jackrabbithanna:
Seems like this has sat long enough, Merging #139

PR #140, opened 2018-02-02
Undefined index: layout in _civicrm_entity_manage_display_submit() at…

yurg:
… line 3908

CiviCRM 4.7.27
Drupal 7.56
CiviCRM Entities  7.x-2.0
PHP	5.6.30


jackrabbithanna:
Thanks for contributing @yurg 
Does the PHP notice you are fixing happen if display suite is not installed?

jackrabbithanna:
Never hurts to check for a variable with isset() ... Merging #140

yurg:
@jackrabbithanna Thank you! re: DisplaySuit : I set up Display Suite for this project only yesterday, but issue appeared 9 days ago, so, perhaps it is not related to DS module.

PR #147, opened 2018-04-19
Field UI & Form improvements

mglaman:
This adds Field UI support. It also re-uses the Node form template and styling that supports a 2/3 and 1/3 column. Where the left side is the main fields and the right side is optional/secondary.

This kicks off Events to have some defaults, see `public static function getFormDisplayInfo($entity_type_id = NULL) {`.

Willing to remove that to bring in the base field definition improvements and Field UI support.


jackrabbithanna:
merging #147

PR #149, opened 2018-06-25
…

stevenvdhout:
Use the getFields method from civicrmApi to pass allong all the field names inside the get method.
Also transforms dates to timestamps and vice versa.

This makes the content inside custom fields from CiviCRM available in Drupal.


jackrabbithanna:
I like this as a beginning to adding custom fields support, but will have to take into consideration multi-valued custom fields at some point.

@mglaman 
Do you think changing the date fields to timestamps breaks anything you add done. Want to take a glance at this?

mglaman:
Converting to timestamps probably simplifies things. Looked on mobile and looks fine.

In currently reviewing how we can support custom Drupal fields, and this looks like there's no conflict. This just improves the Civi API integration

jackrabbithanna:
Merging #149

jackrabbithanna:

PR #150, opened 2018-06-26
Warning: in_array() expects parameter 2 to be array, null given

timdiels1:
When the module is installed or the settings page for Civicrm Entity is viewed, and the settings are not saved yet, you will get a PHP warning:

Warning: in_array() expects parameter 2 to be array, null given in civicrm_entity_entity_type_build() (regel 48 van ../civicrm_entity/civicrm_entity.module)


jackrabbithanna:
Merging #150

jackrabbithanna:

PR #151, opened 2018-06-26
Fix to use config_translations and only show enabled entities in entity reference field

timdiels1:
When trying to use config_translation module, this module produces an error on the routes as it always parses all the entities and not only the enabled ones.

So this commit fixes that only the enabled entities are parsed when the config_translation module is enabled and also only shows the enabled entities to be shown in a entity reference field.


jackrabbithanna:
Merging #151

jackrabbithanna:
pushed to D.O -dev branch

PR #156, opened 2018-07-11
Set ID on New Entity

kerasai:
This change sets the ID on a newly-created entity. Otherwise the ID will be `NULL` until the next time the entity is loaded which causes issues with subsequent operations that utilize the ID.


jackrabbithanna:

PR #157, opened 2018-07-30
Use passed tag_id to find civicrm tag entities

mattwire:
This was a real pain to track down... and I'm not completely sure this is the right way to fix as I'm not really familiar with civicrm_entity module and it's interaction with Drupal. However...

One particular site had created some additional tags (with CiviCRM tag IDs 31,32).  Using the contact "tag" tab it was possible to add them to the contact.
BUT, attempting to remove them caused an exception on the resulting entitytag.delete API call because the ID 31 was being looked up by civicrm_entity as the "entity ID" (ie. the unique primary key for the civicrm_entity_tag table) and not the tag ID.  This meant the call would fail if civicrm_entity_tag.id=31 did not exist, even though that entry linked to a completely different tag ID (civicrm_entity_tag.tag_id=X).

This PR looks up using `tag_id` if entity is "entity_tag" and using `id` otherwise (which was the default before).

I saw a similar specific override for entity_tag at line 3500 of civicrm_entity.module (https://github.com/eileenmcnaughton/civicrm_entity/blob/7.x-2.x/civicrm_entity.module#L3500) which inspired this patch.

### DIFF
--- a/civicrm_entity_controller.inc
+++ b/civicrm_entity_controller.inc
@@ -78,8 +78,19 @@ class CivicrmEntityController extends EntityAPIController {
         }
         if (!empty($ids)) {
           foreach ($ids as $id) {
+            if ($this->entityInfo['description'] == 'entity_tag') {
+              // When removing tags in the civicrm contact tag editor, we get passed the tag ID here, not the entity table ID.
+              $entityParams = [
+                'tag_id' => $id,
+              ];
+            }
+            else {
+              $entityParams = [
+                'id' => $id,
+              ];
+            }
             // we can't rely on civicrm api accepting the 'IN' => array(1,5,6) for all entities
-            $queried_entities += $this->loadEntities(array('id' => $id));
+            $queried_entities += $this->loadEntities($entityParams);
           }
         }
       } catch (Exception $e) {


mattwire:
@pradpnayak could you look at this and see if it seems sensible?

@eileenmcnaughton Am I mad? Does this seem sensible?

eileenmcnaughton:
@jackrabbithanna ping

jackrabbithanna:
Ok thanks for your contribution!  I'll test it out and review it.

jackrabbithanna:
I'm not able to replicate this in 4.6.38 nor 4.7.29 .. will spin up a new CiviCRM 5.x.x test site

You are talking about being able to check or uncheck tags from the "Tags" tab of Contact Summary pages civicrm-admin side?

What version of Civi are you using?

mattwire:
> You are talking about being able to check or uncheck tags from the "Tags" tab of Contact Summary pages civicrm-admin side?

Correct.  It took a long time to track it down to this particular issue!  I think in this case, civicrm_entity is being triggered and hooks firing but no action needs to be taken by civicrm_entity, but because it fails to look up the tag id the parameters get removed before entitytag.delete API can actually delete (uncheck) the tag for the contact.

The site this happened on is running 4.7.31 but has been an issue for quite a while I think.  It was only identified because it happened to be two tags that they just added that they could not uncheck - I now realise this was simply by chance that those new tags have IDs which do not exist in the civicrm_entity_tag table (id, entity_table, entity_id, tag_id).

For example, a tag which has ID 31 and is "checked" for contact ID 123 would be in the civicrm_entity_tag table as:
`XX(auto-increment), civicrm_contact, 123, 31`

But civicrm_entity tries to load this by setting `ID` to 31 instead of tag_id and it will either load the wrong entity or load nothing.  When it loads the wrong entity the "uncheck" a tag works (this can also be reproduced by using the entitytag.delete API directly.  When it loads nothing the "uncheck" a tag fails because loadEntities returns an empty array.

jackrabbithanna:
Ok that makes sense...yeah I can see how that would be a tough one to track down. Thanks for bringing this to our attention.
Merging #157

jackrabbithanna:
pushed to Drupal.org 7.x-2.x-dev . Will be in next release.

jackrabbithanna:
Actually this may be a bug in CiviCRM
It is being triggered by this line:
https://github.com/eileenmcnaughton/civicrm_entity/blob/7.x-2.x/civicrm_entity.module#L3423

the $objectName is EntityTag, but the $id passed is the ID of the tag, not the id of the entity_tag record...

I may have to undo this patch...

jackrabbithanna:
Should hook_civicrm_pre() implementations be getting the id of the tag not the id of the entity_tag, when an entity_tag is deleted ? Is this long standing behavior?

jackrabbithanna:
@mattwire can you try out this PR
https://github.com/eileenmcnaughton/civicrm_entity/pull/158 

Let me know if that fixes it for you.

PR #158, opened 2018-07-30
fix for entity tag delete issue

jackrabbithanna:


mattwire:
@jackrabbithanna This seems to work perfectly!  Many thanks for looking into this.

jackrabbithanna:
Ok merging #158 and pushing to Drupal.org

PR #159, opened 2018-08-01
FieldConfig, Views integration, various bug fixes.

mglaman:


civicrm_entity.module
```diff
@@ -44,34 +53,38 @@ function civicrm_entity_entity_type_build(array &$entity_types) {
...
-        'class' => \Drupal\civicrm_entity\Entity\CivicrmEntity::class,
-        'originalClass' => \Drupal\civicrm_entity\Entity\CivicrmEntity::class,
+        'class' => CivicrmEntity::class,
+        'originalClass' => CivicrmEntity::class,
         'id' => $entity_type_id,
         'civicrm_entity' => $civicrm_entity_name,
         'civicrm_entity_ui_exposed' => in_array($entity_type_id, $enabled_entity_types),
-        'label' => new TranslatableMarkup('CiviCRM :name', [':name' => $civicrm_entity_info['civicrm entity label']]),
+        'civicrm_required_fields' => !empty($civicrm_entity_info['required']) ? $civicrm_entity_info['required'] : [],
+      'label' => new TranslatableMarkup('CiviCRM :name', [':name' => $civicrm_entity_info['civicrm entity label']]),
```
mglaman: Nit: spacing

civicrm_entity.module
```diff
@@ -44,34 +53,38 @@ function civicrm_entity_entity_type_build(array &$entity_types) {
...
         // @todo add label_singular
         // @todo add label_plural
         // @todo add label_count
         'entity_keys' => [
           'id' => 'id',
           'label' => $civicrm_entity_info['label property'],
         ],
-        'admin_permission' => 'administer civicrm entity',
+        'base_table' => $entity_type_id,
+      'admin_permission' => 'administer civicrm entity',
```
mglaman: Nit: spacing

civicrm_entity.module
```diff
@@ -93,30 +106,19 @@ function civicrm_entity_entity_type_build(array &$entity_types) {
  * Provides the pseudoconstant values for CiviCRM entity fields.
  *
  * @param \Drupal\Core\Field\FieldStorageDefinitionInterface $definition
- *  The field storage definition.
- * @param \Drupal\Core\Entity\FieldableEntityInterface|NULL $entity
- *  The entity.
+ *   The field storage definition.
+ * @param \Drupal\Core\Entity\FieldableEntityInterface|null $entity
```
mglaman: nit: s/null/NULL

src/CivicrmEntityListBuilder.php
```diff
@@ -23,9 +26,31 @@ public function buildHeader() {
...
+    ];
+
+    return $operations;
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  protected function ensureDestination(Url $url) {
+    // @todo Disabled redirection to collection until paging fixed.
```
mglaman: I believe it was fixed and we can remove this.

src/Entity/CivicrmEntity.php
```diff
@@ -17,13 +22,98 @@
...
     $fields = [];
+    $civicrm_entity_info = SupportedEntities::getInfo()[$entity_type->id()];
+    $civicrm_required_fields = $entity_type->get('civicrm_required_fields');
     $field_definition_provider = \Drupal::service('civicrm_entity.field_definition_provider');
-    $civicrm_fields = \Drupal::service('civicrm_entity.api')->getFields($entity_type->get('civicrm_entity'));
+    $civicrm_fields = \Drupal::service('civicrm_entity.api')->getFields($entity_type->get('civicrm_entity'), 'create');
     foreach ($civicrm_fields as $civicrm_field) {
+      // If the entity has any field overrides, merge them in.
+      if (!empty($civicrm_entity_info['fields'][$civicrm_field['name']])) {
+        $civicrm_field = $civicrm_entity_info['fields'][$civicrm_field['name']] + $civicrm_field;
```
mglaman: This actually doesn't merge, but just fills missing values. Comment is misleading.

mglaman (commented):
Initial nits and things.

dsnopek:
Thanks @mglaman! Like you, I'll be going through and doing line-by-line code review later, so, maintainers, don't merge yet! :-)

jackrabbithanna:
Hell yeah this looks like the big commit! Holding on merge for until @dsnopek releases it...

Big thanks @mglaman and @dsnopek

civicrm_entity.module
```diff
@@ -44,34 +53,38 @@ function civicrm_entity_entity_type_build(array &$entity_types) {
...
-            'add' => \Drupal\civicrm_entity\Form\CivicrmEntityForm::class,
-            'edit' => \Drupal\civicrm_entity\Form\CivicrmEntityForm::class,
-            'delete' => \Drupal\Core\Entity\ContentEntityDeleteForm::class,
+            'default' => CivicrmEntityForm::class,
+            'add' => CivicrmEntityForm::class,
+            'edit' => CivicrmEntityForm::class,
+            'delete' => ContentEntityDeleteForm::class,
           ],
+        'views_data' => CivicrmEntityViewsData::class,
+        'storage_schema' => CivicrmEntityStorageSchema::class,
```
dsnopek: Nit: spacing

src/CiviEntityStorage.php
```diff
@@ -246,5 +375,248 @@ protected function initFieldValues(ContentEntityInterface $entity, array $values
...
+        }
+      }
+    }
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function requiresEntityStorageSchemaChanges(EntityTypeInterface $entity_type, EntityTypeInterface $original) {
+    // There is no base table.
```
dsnopek: This comment isn't exactly true. There _is_ a base table, but we don't manage it in Drupal

src/CiviEntityStorage.php
```diff
@@ -246,5 +375,248 @@ protected function initFieldValues(ContentEntityInterface $entity, array $values
...
+    return FALSE;
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function requiresEntityDataMigration(EntityTypeInterface $entity_type, EntityTypeInterface $original) {
+    // There is no base table.
```
dsnopek: Same here

src/CivicrmEntityListBuilder.php
dsnopek: The destination is still not the most usable - I think there's more we could do here. But that doesn't mean we need a `@todo` :-)

src/Entity/CivicrmEntity.php
```diff
@@ -17,13 +22,98 @@
  */
 class CivicrmEntity extends ContentEntityBase {
 
+  /**
+   * {@inheritdoc}
+   */
   public static function baseFieldDefinitions(EntityTypeInterface $entity_type) {
     $fields = [];
+    $civicrm_entity_info = SupportedEntities::getInfo()[$entity_type->id()];
+    $civicrm_required_fields = $entity_type->get('civicrm_required_fields');
```
dsnopek: Isn't the `$entity_type->get('civicrm_required_fields')` derived from info from `SupportedEntities::getInfo()`? I wonder if we could just stash all the info from `SupportedEntities::getInfo()` on the entity type so we don't have to keep calling it? That would also mean that developers would have a place (ie. `hook_entity_type_alter()`) to alter this info if they wanted.

Not super important to this PR, just a thought :-)

src/Entity/Sql/CivicrmEntityStorageSchema.php
```diff
@@ -0,0 +1,93 @@
...
+   * {@inheritdoc}
+   */
+  public function requiresEntityStorageSchemaChanges(EntityTypeInterface $entity_type, EntityTypeInterface $original) {
+    return FALSE;
+  }
+
+  /**
+   * {@inheritdoc}
+   *
+   * There are no base tables.
```
dsnopek: Like the other comments about base tables noted above.

src/Plugin/EntityReferenceSelection/CivicrmContactSelection.php
```diff
@@ -0,0 +1,37 @@
...
+use Drupal\Component\Utility\Html;
+use Drupal\Core\Entity\Plugin\EntityReferenceSelection\DefaultSelection;
+use Drupal\Core\Form\FormStateInterface;
+use Drupal\taxonomy\Entity\Vocabulary;
+
+/**
+ * Provides specific access control for the civicrm_contact entity type.
+ *
+ * This is because `display_name` does not work on LIKE queries, but the
+ * `=` condition is treated as one.
```
dsnopek, line 14: Is this behavior only for display_name? If so, we should probably check the name of the field being matched and swap LIKE for '='. Otherwise, this could cause problems if the user chooses a different field to autocomplete on.

src/Plugin/Field/FieldWidget/TextareaWidget.php
```diff
@@ -0,0 +1,44 @@
...
+      $config->get('filter_format') ?: filter_fallback_format(),
+    ];
+    return $element;
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public static function isApplicable(FieldDefinitionInterface $field_definition) {
+    return strpos($field_definition->getTargetEntityTypeId(), 'civicrm_') !== FALSE;
```
dsnopek: Hm. I wonder if this should also check if it's a base field? Because we don't really need this widget on configurable fields, just the base fields from CiviCRM.

dsnopek:
@mglaman This all looks awesome! Nothing I found was critical.

@jackrabbithanna Thanks :-) FYI, the main architectural issue that we kept coming back to with these changes was if we should extend `SqlContentEntityStorage` because far too much code targets that class specifically (rather than `SqlEntityStorageInterface` or `TableMappingInterface`). We tried it a couple times and it turned out to be unworkable due to some issues in Drupal's Entity API outlined on this meta issue:

https://www.drupal.org/project/drupal/issues/2960147

Once that issue is fixed, I think we could extend `SqlContentEntityStorage` and remove the copy-paste code that we took from core. But all of the copy-paste code is clearly marked, so it should be easy to update and/or eliminate in the future. There's been some recent action on the child issues of the above issue, because it's blocking some other core work, namely the Workflow Initiative, so it's possible that it could be fixed by Drupal 8.7. Finger's crossed!

Anyway, once all the little things commented above are fixed, this should be ready to merge in my opinion :-)

src/Entity/CivicrmEntity.php
mglaman: So I was afraid of stashing all of the info from `SupportedEntities::getInfo()` on the entity annotation, because it then becomes cached and added into memory. Which could bloat performance pretty quick.

I'd rather remove this `get` and just pull from `getInfo`, allowing end users to just modify the entity definitions on their own (because that is an edge case to further modify.)

src/Plugin/EntityReferenceSelection/CivicrmContactSelection.php
mglaman: `display_name` is set to be the label, which is the default reference. 

```
    if (isset($match) && $label_key = $entity_type->getKey('label')) {
      $query->condition($label_key, $match, $match_operator);
    }
```

The label is always how the selection plugins work, unless a View is used.

mglaman:
@dsnopek addressed your initial items. For entity reference fields the only configuration for fields is sorting. Otherwise `match` is always against the label property. Now, if someone altered the definition it could break. But I think that is an edge case.

src/Plugin/EntityReferenceSelection/CivicrmContactSelection.php
dsnopek: Ah, ok, understood! Thanks :-)

dsnopek:
Thanks, @mglaman!

@jackrabbithanna This is ready to be considered for merge! There shouldn't be any functional changes since Matt originally posted the PR. I don't know if you've had a chance to test it at all? We've been testing it with some advanced events functionality for a customer site (not live yet), and it's been working pretty great in that testing :-)

jackrabbithanna:
@dsnopek I'm going to merge this, and if any lil issues arise, we can focus on them one by one. I want to get this big set of work in the project and available to people to start using / testing / providing feedback. 

I haven't tested it yet, but I will after this.  What are you thoughts about releasing this as an "alpha1" ? Any reason not to? Also, I'd like to do some marketing, that is let people know the project status, all the work that is being done etc....

Since you've been primary on this development for D8, I want to give you the chance to say anything you want to say, or write your own blog post. Regardless I'll write a short post of my own to make people aware. We can discuss this here or take it to Mattermost if you like.

jackrabbithanna:
Pushed to 8.x-3.x-dev on D.O project page

dsnopek:
Woohoo! Thanks :-)

We have some little things we want to get in in the next couple days / weeks. But I don't think we're going to be doing anything else super big in the near-term, this is already working enough to accomplish quite a lot. :-)

So, from my perspective, doing an alpha1 release with the current code would be fine! It'd be great to do alpha2 in a couple weeks too, after some of this little stuff, but yeah - I think this is a great time to get other folks looking at it.

Regarding marketing: I think we'd probably want to write our own blog post. We can coordinate all that a a little if you'd like.

Thanks!

PR #160, opened 2018-08-06
use number_format() on price set field registration form for amounts

jackrabbithanna:
format price set options to two decimal places (something changed in 5.x as price field value amounts now return many decimal places)


jackrabbithanna:
Pushed to D.O 7.x-2.x-dev

PR #161, opened 2018-08-07
Fix fatal error if 'api/v3/utils.php' isn't loaded

dsnopek:
Just a little fix!


jackrabbithanna:
Merging #161

jackrabbithanna:
Pushed to 8.x-3.x-dev on D.O project page

PR #162, opened 2018-08-08
Make civicrm entities themable and add a default civicrm-entity template

shadcn:
CiviCRM entities are not themable so you cannot render them in a theme. This PR uses a ViewBuilder to add a default theme for CiviCRM entities and make them themable.

### DIFF
--- a/civicrm_entity.module
+++ b/civicrm_entity.module
@@ -149,7 +149,22 @@ function template_preprocess_civicrm_entity(&$variables) {
  * Implements hook_theme_suggestions_HOOK_alter().
  */
 function civicrm_entity_theme_suggestions_civicrm_entity_alter(array &$suggestions, array $variables) {
-  $entity_type = $variables['elements']['#entity_type'];
+  $entity_type = NULL;
+  if (isset($variables['elements']['#entity_type'])) {
+    $entity_type = $variables['elements']['#entity_type'];
+  }
+
+  // Find the CivicrmEntity from elements if #entity_type is not set.
+  if (! $entity_type) {
+    foreach ($variables['elements'] as $element) {
+      if ($element instanceof CivicrmEntity) {
+        /** @var CivicrmEntity $element */
+        $entity_type = $element->getEntityTypeId();
+        break;
+      }
+    }
+  }
+
   $view_mode = $variables['elements']['#view_mode'];
   $hook = $variables['theme_hook_original'];
 
--- a/civicrm_entity.module
+++ b/civicrm_entity.module
@@ -142,35 +142,40 @@ function template_preprocess_civicrm_entity(&$variables) {
   $variables['view_mode'] = $variables['elements']['#view_mode'];
 
   // Add the bundle to the template.
-  $variables['entity_type'] = $variables['elements']['#entity_type'];
+  $variables['entity_type'] = _civicrm_entity_get_entity_type_from_elements($variables['elements']);
 }
 
 /**
  * Implements hook_theme_suggestions_HOOK_alter().
  */
 function civicrm_entity_theme_suggestions_civicrm_entity_alter(array &$suggestions, array $variables) {
-  $entity_type = NULL;
-  if (isset($variables['elements']['#entity_type'])) {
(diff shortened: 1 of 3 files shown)


src/CiviCrmEntityViewBuilder.php
```diff
@@ -0,0 +1,22 @@
...
+class CiviCrmEntityViewBuilder extends EntityViewBuilder {
+
+  /**
+   * {@inheritdoc}
+   */
+  protected function getBuildDefaults(EntityInterface $entity, $view_mode) {
+    $defaults = parent::getBuildDefaults($entity, $view_mode);
+
+    // Set a default theme.
+    $defaults['#theme'] = 'civicrm_entity';
```
mglaman, line 17: This is required because we're not generating theme hooks (and ergo template files) for each discoverable entity type. The parent logic contains

```
    // Add the default #theme key if a template exists for it.
    if ($this->themeRegistry->getRuntime()->has($this->entityTypeId)) {
      $build['#theme'] = $this->entityTypeId;
    }
```

So +1 to this, it standardizes the theme hook for all CiviCRM entities.

civicrm_entity.module
```diff
@@ -121,3 +127,35 @@ function civicrm_entity_pseudoconstant_options(FieldStorageDefinitionInterface $
...
+/**
+ * Implements hook_theme_suggestions_HOOK_alter().
+ */
+function civicrm_entity_theme_suggestions_civicrm_entity_alter(array &$suggestions, array $variables) {
+  $entity_type = $variables['elements']['#entity_type'];
+  $view_mode = $variables['elements']['#view_mode'];
+  $hook = $variables['theme_hook_original'];
+
+  // Add a suggestion based on the entity type.
+  $suggestions[] = $hook . '__' . $entity_type;
```
mglaman: Only "nit" would be to explain someone could theme events by:

`civicrm-entity--civicrm-event.html.twig`

mglaman: But maybe that belongs in the twig file.

mglaman (approved):
This looks great!

civicrm_entity.module
jackrabbithanna: Theme file makes sense to me, that's where the suggestions have been commented in D7 templates...

src/CiviCrmEntityViewBuilder.php
jackrabbithanna: I like this too, wish I had done it this way for D7....one standard file, that is over-ridable per entity type by template name..less files to maintain. Might just "backport" this.

civicrm_entity.module
shadcn: Hmm I found a bug where civicrm_contact does not have the `#entity_type` key. We might need to loop through `$variables['elements']` and find the `CivicrmEntity`. Or maybe there's a better way?

screenshot

shadcn: The commit below fixes this.

templates/civicrm-entity.html.twig
```diff
@@ -0,0 +1,16 @@
+{#
+/**
+ * @file
+ * Default template for a CiviCRM template.
+ */
```
jackrabbithanna: Perhaps some template naming suggestions documented here for how to make a template for one entity type

dsnopek:
I just tested this and it works great! :+1:

templates/civicrm-entity.html.twig
dsnopek: I concur! I had to look at the code in `hook_theme_suggestions_HOOK_alter()` to figure out how to name the template file. Having it documented somewhere would be good.

@arshad

shadcn: @dsnopek Done. I've updated the comment in `civicrm-entity.html.twig`.

dsnopek: Looks great, thanks!

dsnopek:
@jackrabbithanna I think this is ready for another look!

jackrabbithanna:
Ok great, Merging #162 and pushing this and all merged PRs to drupal.org project page, 8.x-3.x-dev

PR #166, opened 2018-08-15
Allow Edits in CiviCRM to Invoke Drupal Entity CRUD Hooks

kerasai:
We've been noticing that working with the CiviCRM data in the Civi administration was causing odd issues (primarily caching with Views and in Menus) with the Drupal entity handling.

This change adds `civicrm_pre` and `civicrm_post` hook implementations which invoke the appropriate Drupal entity CRUD hooks.

**Implementation Notes**

In the Drupal world, these entity hooks are typically invoked by the entity storage service. I've added public methods to the `CiviEntityStorage` which allow the outside code (the CiviCRM hooks) to initiate the invoking of the hooks, through the entity storage service.

I also added a utility method `SupportedEntities::getDrupalType` which maps the camel-cased CiviCRM object names to the snake-cased Drupal entity types.

**TODO:**

- [x] Investigate methods of preventing the the CiviCRM hooks from invoking the Drupal hooks when the save is initiated by Drupal (invokes Drupal hooks twice in this case)
- [x] Investigate how to obtain first-class Drupal entities in the CiviCRM hooks, before passing to the storage handler to run the Drupal hooks (at this point, they're kind of piecemealed from available data)
- [ ] General sanity checking

### DIFF
--- a/civicrm_entity.module
+++ b/civicrm_entity.module
@@ -192,9 +192,7 @@ function civicrm_entity_civicrm_pre($op, $objectName, $id, &$params) {
   $entityType = SupportedEntities::getEntityType($objectName);
 
   // Check valid entity type.
-  $config = \Drupal::config('civicrm_entity.settings');
-  $enabled_entity_types = $config->get('enabled_entity_types') ?: [];
-  if (!in_array($entityType, $enabled_entity_types)) {
+  if (!$entityType) {
     return;
   }
 
@@ -243,9 +241,7 @@ function civicrm_entity_civicrm_post($op, $objectName, $objectId, &$objectRef) {
   $entityType = SupportedEntities::getEntityType($objectName);
 
   // Check valid entity type.
-  $config = \Drupal::config('civicrm_entity.settings');
-  $enabled_entity_types = $config->get('enabled_entity_types') ?: [];
-  if (!in_array($entityType, $enabled_entity_types)) {
+  if (!$entityType) {
     return;
   }
 
--- a/civicrm_entity.module
+++ b/civicrm_entity.module
@@ -189,9 +189,14 @@ function civicrm_entity_civicrm_pre($op, $objectName, $id, &$params) {
     return;
   }
 
-  /** @var \Drupal\civicrm_entity\CiviCrmApiInterface $api */
-  $api = \Drupal::service('civicrm_entity.api');
-  $entityType = 'civicrm_' . $api->getEntityNameFromCamel($objectName);
+  $entityType = SupportedEntities::getEntityType($objectName);
+
+  // Check valid entity type.
+  $config = \Drupal::config('civicrm_entity.settings');
+  $enabled_entity_types = $config->get('enabled_entity_types') ?: [];
+  if (!in_array($entityType, $enabled_entity_types)) {
+    return;
+  }
 
   /** @var \Drupal\civicrm_entity\CiviEntityStorage $storage */
   $storage = \Drupal::entityTypeManager()->getStorage($entityType);
@@ -226,9 +231,14 @@ function civicrm_entity_civicrm_post($op, $objectName, $objectId, &$objectRef) {
     return;
   }
 
-  /** @var \Drupal\civicrm_entity\CiviCrmApiInterface $api */
-  $api = \Drupal::service('civicrm_entity.api');
-  $entityType = 'civicrm_' . $api->getEntityNameFromCamel($objectName);
+  $entityType = SupportedEntities::getEntityType($objectName);
+
+  // Check valid entity type.
+  $config = \Drupal::config('civicrm_entity.settings');
+  $enabled_entity_types = $config->get('enabled_entity_types') ?: [];
+  if (!in_array($entityType, $enabled_entity_types)) {
+    return;
+  }
 
   /** @var \Drupal\civicrm_entity\CiviEntityStorage $storage */
   $storage = \Drupal::entityTypeManager()->getStorage($entityType);
--- a/civicrm_entity.module
+++ b/civicrm_entity.module
@@ -184,7 +184,7 @@ function _civicrm_entity_get_entity_type_from_elements($elements) {
  * Implements hook_civicrm_pre().
  */
 function civicrm_entity_civicrm_pre($op, $objectName, $id, &$params) {
-  $operations = ['create', 'edit', 'delete'];
+  $operations = ['create', 'edit', 'delete', 'restore'];
   if (!in_array($op, $operations)) {
     return;
   }
@@ -221,6 +221,10 @@ function civicrm_entity_civicrm_pre($op, $objectName, $id, &$params) {
       $storage->civiPreDelete($entity);
       break;
 
+    case 'restore':
+      $storage->civiPreSave($entity);
+      break;
+
     case 'edit':
       $storage->civiPreSave($entity);
       break;
@@ -231,8 +235,7 @@ function civicrm_entity_civicrm_pre($op, $objectName, $id, &$params) {
  * Implements hook_civicrm_post().
  */
 function civicrm_entity_civicrm_post($op, $objectName, $objectId, &$objectRef) {
-  // TODO: add restore.
-  $operations = ['create', 'edit', 'delete'];
+  $operations = ['create', 'edit', 'delete', 'restore'];
   if (!in_array($op, $operations)) {
     return;
   }
@@ -257,6 +260,10 @@ function civicrm_entity_civicrm_post($op, $objectName, $objectId, &$objectRef) {
 
     case 'delete':
       $storage->civiPostDelete($entity);
+      break;
+
+    case 'restore':
+      $storage->civiPostSave($entity, TRUE);
       break;
 
     case 'edit':
--- a/civicrm_entity.module
+++ b/civicrm_entity.module
@@ -218,8 +218,10 @@ function civicrm_entity_civicrm_pre($op, $objectName, $id, &$params) {
  * Implements hook_civicrm_post().
  */
 function civicrm_entity_civicrm_post($op, $objectName, $objectId, &$objectRef) {
+  /** @var \Drupal\civicrm_entity\CiviCrmApiInterface $api */
+  $api = \Drupal::service('civicrm_entity.api');
   $operations = ['create', 'update', 'delete'];
-  if (!in_array($op, $operations) || !$entityType = SupportedEntities::getDrupalType($objectName)) {
+  if (!in_array($op, $operations) || !$entityType = $api->getEntityNameFromCamel($objectName)) {
     return;
   }
 
--- a/civicrm_entity.module
+++ b/civicrm_entity.module
@@ -200,10 +200,15 @@ function civicrm_entity_civicrm_pre($op, $objectName, $id, &$params) {
 
   /** @var \Drupal\civicrm_entity\CiviEntityStorage $storage */
   $storage = \Drupal::entityTypeManager()->getStorage($entityType);
+
   if ($op == 'create') {
     $entity = $storage->create($params);
   }
   else {
+    // Special handling for EntityTag objects.
+    if ($objectName == 'EntityTag') {
+      $id = $storage->getEntityTagEntityId($params[0][0], $params[1]);
+    }
     $entity = $storage->load($id);
   }
 
@@ -226,6 +231,7 @@ function civicrm_entity_civicrm_pre($op, $objectName, $id, &$params) {
  * Implements hook_civicrm_post().
  */
 function civicrm_entity_civicrm_post($op, $objectName, $objectId, &$objectRef) {
+  // TODO: add restore.
   $operations = ['create', 'edit', 'delete'];
   if (!in_array($op, $operations)) {
     return;
--- a/civicrm_entity.module
+++ b/civicrm_entity.module
@@ -184,8 +184,10 @@ function _civicrm_entity_get_entity_type_from_elements($elements) {
  * Implements hook_civicrm_pre().
  */
 function civicrm_entity_civicrm_pre($op, $objectName, $id, &$params) {
+  /** @var \Drupal\civicrm_entity\CiviCrmApiInterface $api */
+  $api = \Drupal::service('civicrm_entity.api');
   $operations = ['create', 'update', 'delete'];
-  if (!in_array($op, $operations) || !$entityType = SupportedEntities::getDrupalType($objectName)) {
+  if (!in_array($op, $operations) || !$entityType = $api->getEntityNameFromCamel($objectName)) {
     return;
   }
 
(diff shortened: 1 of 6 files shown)


civicrm_entity.module
```diff
@@ -179,3 +179,66 @@ function _civicrm_entity_get_entity_type_from_elements($elements) {
...
+function civicrm_entity_civicrm_pre($op, $objectName, $id, &$params) {
+  $operations = ['create', 'update', 'delete'];
+  if (!in_array($op, $operations) || !$entityType = SupportedEntities::getDrupalType($objectName)) {
+    return;
+  }
+
+  /** @var \Drupal\civicrm_entity\CiviEntityStorage $storage */
+  $storage = \Drupal::entityTypeManager()->getStorage($entityType);
+  $entity = $storage->create($params);
+  if ($id) {
```
kerasai: Delete operations don't provide parameters, so we have to at least set an ID. Maybe a better solution would be to load the full Drupal entity for `edit` and `delete` operations.

src/CiviEntityStorage.php
```diff
@@ -619,4 +620,60 @@ protected function saveToDedicatedTables(ContentEntityInterface $entity, $update
...
+   *
+   * @param \Drupal\Core\Entity\EntityInterface $entity
+   *   The entity to be deleted.
+   *
+   * @see \Drupal\Core\Entity\EntityStorageInterface::delete
+   */
+  public function civiPreDelete(EntityInterface $entity) {
+    CivicrmEntity::preDelete($this, [$entity]);
+    $this->invokeHook('predelete', $entity);
+  }
```
kerasai: The base entity storage classes don't have a clean pre/post delete handling (like the pre/post save) so these `::civiPreDelete` and `::civiPostDelete` methods have to do a bit more work than just passing off to existing code.

jackrabbithanna:
FYI, the EntityTag entity is weird! In a hook_civicrm_post() hook the id given will be the value of tag_id not id of the entity tag record...
Just had to fix a bug in the 7.x version regarding this: 
https://github.com/eileenmcnaughton/civicrm_entity/pull/158

civicrm_entity.module
dsnopek: Loading the full Drupal entity for 'edit' and 'delete' makes sense, because the Drupal hooks are probably expecting that anyway. In fact, I'd be a little worried that not doing that could create weird interactions with certain hook implementations down the line. For example, some hook expecting the entity to be in a particular state after being loaded, but it isn't, because the entity wasn't actually loaded.

src/SupportedEntities.php
```diff
@@ -688,4 +690,24 @@ public static function getFormDisplayInfo($entity_type_id = NULL) {
...
+   *
+   * @param $objectName
+   *   The CiviCRM object name.
+   *
+   * @return string|null
+   *   The Drupal entity type, or FALSE if not defined.
+   */
+  public static function getDrupalType($objectName) {
+    $converter = new CamelCaseToSnakeCaseNameConverter();
+    $entityType = 'civicrm_' . $converter->normalize($objectName);
```
dsnopek: This is assuming the Drupal entity name is always "civicrm_" plus the snake cased Drupal entity name. Since we're specifying the "civicrm entity name" on the info array, we should actually be using that. This method could statically cache a lookup array built from looping over the info array?

civicrm_entity.module
```diff
@@ -179,3 +179,66 @@ function _civicrm_entity_get_entity_type_from_elements($elements) {
 
   return NULL;
 }
+
+/**
+ * Implements hook_civicrm_pre().
+ */
+function civicrm_entity_civicrm_pre($op, $objectName, $id, &$params) {
+  $operations = ['create', 'update', 'delete'];
+  if (!in_array($op, $operations) || !$entityType = SupportedEntities::getDrupalType($objectName)) {
```
dsnopek: I don't know if this actually violates Drupal style or not, but this bit seems ambiguous for humans: `!$entityType = SupportedEntities::getDrupalType($objectName)`

I think it'd be better to wrap the assignment in parenthesis and negate that.

dsnopek:
Re:

> Investigate methods of preventing the the CiviCRM hooks from invoking the Drupal hooks when the save is initiated by Drupal (invokes Drupal hooks twice in this case)

This could be a static array in the `civi(Pre|Post)(Save|Delete)` methods that tracks the entity type and id, and returns without doing anything if it's present?

You'd need to be sure to wrap this in a `try { ... } finally { ... }` so that the item gets removed from the array before exiting the method, even if there's an exception thrown.

civicrm_entity.module
jackrabbithanna: I would prefer to load the Drupal version of the entity for sure....If you notice in the 7.x version, in civicrm_entity_civicrm_post(), that is where we trigger Rules events, and you'll notice we did NOT load the object there....this is because of problems with multi-valued custom fields....The entity_metadata_wrapper in D7 would load a wrapper from the $objectRef, but it's not ideal, and has caused a few problems in the past..

jackrabbithanna:
> Investigate how to obtain first-class Drupal entities in the CiviCRM hooks, before passing to the storage handler to run the Drupal hooks (at this point, they're kind of piecemealed from available data)

In the hook_civicrm_post() you should be able to just load the Drupal entity from the ID in the hooks....the drupal entity controller should do a fresh API call at this point,  and you'll get your custom field values etc...At least the "latest" value...multi-valued custom field groups throw a bit of a monkey wrench into this. It would be a good idea as you develop this, to put a multi-valued custom field group on an entity like Contact, for testing...Handling multi-valued custom field groups has always been tricky. 

In hook_civicrm_pre() that may not be the case though, because a load of the entity could load "old" data...In the 7.x version, we only used the _pre hook to trigger field_attach_delete() to remove any field data before the entity was deleted in CiviCRM...we loaded the Drupal entity, and it didn't really matter much if the data was stale..because it was about to be deleted...

We never invoked hook_entity_presave() implementations in the 7.x version....Although it should be possible to construct the entity object, pass it to hook implementations, get the values back, and set corresponding values in the $params array, but you may not be able to alter custom field values in the presave() hook implementations...

As far as I know, there is no "pre" hook for civicrm custom field data...hook_civicrm_custom() implementations are called AFTER the database write.

kerasai:
I've made a number of additional changes, mostly focused on re-using the existing D7 logic and code where appropriate.

I think this is ready for another look.

The one thing I've yet to address is the duplicate calls to the Drupal hooks, when saved from Drupal. I don't think the solution proposed by @dsnopek is sufficient, because the first time through it's not going through our `civi(Pre/Post)(Save/Delete)` methods.

I think a better solution may to have the Drupal entity contain a marker that indicates if it's in the process of a CRUD operation. This would be set and cleared by the entity storage service, and the Civi hooks would bail if they found the operation was initiated by Drupal. Thoughts?

dsnopek:
Ah, right. :-/ I think your proposed solution using a property on the Drupal entity makes sense, we just have to make sure everything gets cleared so saving the same entity twice still triggers the hooks twice (just not four times).

civicrm_entity.module
```diff
@@ -179,3 +179,124 @@ function _civicrm_entity_get_entity_type_from_elements($elements) {
...
+    return;
+  }
+
+  $entityType = SupportedEntities::getEntityType($objectName);
+
+  // Check valid entity type.
+  $config = \Drupal::config('civicrm_entity.settings');
+  $enabled_entity_types = $config->get('enabled_entity_types') ?: [];
+  if (!in_array($entityType, $enabled_entity_types)) {
+    return;
```
dsnopek: I'm not 100% on this, but I think we want the hooks to fire even if the entity isn't "enabled". When Matt and I originally discussed the enabled/disabled thing we came to the conclusion that it should just hide/show the entity type in the UI, because making the underlying entity stuff behave differently makes for a very weird, undeclarable dependency. But this is a little different than the issue we originally discussed...

kerasai: My main concern was that we just assume a Drupal entity type exists, then start using the storage service, etc. There are some things that are (or at least via the API appear to be) Civi objects but aren't defined for use in this module's `SupportedEntities` class.

Perhaps just call `SupportedEntities` to see if it's defined there?

jackrabbithanna: @dsnopek your last comment would be a departure from the D7 functionality, where if an entity is not enabled, then it would NOT engage any Drupal systems. The original idea behind that was basically a performance and permissions access issue, because over 40 entity types are exposed to Drupal. If an entity was disabled, then it would not be included in civicrm_entity_entity_info, and would be unavailable for any Drupal API integration...You do have a good point though about the undeclared dependency issue though...because for instance if a Rule had been created to respond to a certain entity CRUD operation, and then later an admin disables an entity type, then you have a broken Rule left hanging...Or if a custom module manipulated a civicrm entity type and then all of a sudden it's not exposed anymore you could conceivably get a broken site...Basically it was assumed that it is the module developers or site admins responsibility to keep track of or manage what is enabled. What I did for some of the submodules of the D7 version was on install/enable, if a entity type was required, it would enable the entity, and disable the checkbox element on the settings form so that required entity types could NOT be disabled via the UI, while the submodule was enabled. Can you lay out the case a little more for why the Drupal hooks should be called even if an entity type is disabled? Maybe there are other reasons, or D8 specific reasons I am not thinking of as well? I'll think about it some more too, you are right that it is a complex dependency to understand for people developing modules that use CiviCRM Entity. I'm also seeing how people could have trouble with broken Views if some admin disabled an entity type, and Views integration for that entity type goes away.

dsnopek: Yeah, I think this is something we need to do differently for D8. In D8, there's a number of places where you're required to declare the dependencies something has in order for everything to work correctly. For example, every config object in CMI needs to say what it depends on. If a bit of config needs a particular entity type from civicrm_entity, it can depend on the civicrm_entity module, but it can't depend on civicrm_entity plus that the fact that the entity type is enabled, because that is just a specific property within another config object (ie. we can't declare a dependency on that).

Side note: if civicrm_entity types were enabled by creating an entity, this would be different, because a config object can depend on an entity, but do we really want a whole entity type just to enable entity types? That sounds a little crazy.

So, the simple solution is to say the entity types always exist, but we then we try to hide the disabled ones to prevent exploding the UI.

jackrabbithanna: Ok thanks for your patience and explanation, let's not get crazy by using an entity type to enable entity types :) I think what we could do in the future is offer developers a way to exclude/include supported entity types. We could add an _alter() hook in getInfo() method of the SupportedEntities class....or we turn that class into a service and people could use service decorators, or maybe supported entities becomes a plugin type. It will become an issue down the road when people want to be able to integrate their custom entity types, and alter default properties of supported entities.

dsnopek: Re @kerasai: 

> Perhaps just call `SupportedEntities` to see if it's defined there?

Yeah, I think that's the way to go.

Re @jackrabbithanna:

> I think what we could do in the future is offer developers a way to exclude/include supported entity types. We could add an _alter() hook in getInfo() method of the SupportedEntities class....or we turn that class into a service and people could use service decorators, or maybe supported entities becomes a plugin type.

I concur, we need some mechanism for this.

On one of Matt's earlier PR's I put forward the idea that the `SupportedEntities::getInfo()` information gets transferred straight to the entity types and then we look only at the entity types, so that other modules could alter the data (there's already an alter hook for entity types).

But having our own alter hook or using service decorators or maybe an event, all sound like fine ideas too. I'm kind of skeptical of using a plugin type, since the entity types can't be declared arbitrarily (the civicrm entity type needs to already exist), but there is some precedent in contrib for using plugin types like that (ie. more like 'handlers' than plugins).

jackrabbithanna: @kerasai For sure checking in SupportedEntities is the way to go here, since we will be standardizing this way for determining what entities are exposed as Drupal entities. If anywhere else in the code base, the check should be made from there too. @dsnopek youll think though excepting manage fields forms, manage display forms, or drupal native pages / forms?  . I can imagine developers wanting to at least add fields and manage them, if CiviCRM data was used by the REST API, or for other programmatic uses, say in these hooks....perhaps only Manage Display and the view page / edit forms should be excluded when disabled via the config page.

kerasai:
I think this is all set for a full on review and merge if acceptable.

I have made the following changes:

- Adjusted logic in `SupportedEntities` to determine if the entity type was valid for handling
- Added the `::original` property to the entity passed to the Drupal hooks, as things were breaking as some modules attempt to utilize it in `entity_*` hooks
- Added a `drupal_crud` field to the entity as a marker to indicate if Drupal initiated the save
- Overrode `::save()` and `::delete()` methods on the entity class to set and clear the `drupal_crud` field

src/Entity/CivicrmEntity.php
```diff
@@ -21,6 +22,27 @@
...
 class CivicrmEntity extends ContentEntityBase {
 
+  /**
+   * {@inheritdoc}
+   */
+  public function save() {
+    // Set ::drupal_crud to indicate save is coming from Drupal.
+    $this->drupal_crud = TRUE;
+    $result = parent::save();
+    $this->drupal_crud = FALSE;
```
dsnopek: I think this should probably have a `try { ... } finally { ... }` to make sure that `$this->drupal_crud = FALSE;` is run even if an exception is thrown while saving

src/Entity/CivicrmEntity.php
```diff
@@ -21,6 +22,27 @@
...
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function delete() {
+    // Set ::drupal_crud to indicate delete is coming from Drupal.
+    $this->drupal_crud = TRUE;
+    parent::delete();
+    $this->drupal_crud = FALSE;
```
dsnopek: This one too

src/Entity/CivicrmEntity.php
```diff
@@ -39,6 +61,7 @@ public static function baseFieldDefinitions(EntityTypeInterface $entity_type) {
       $fields[$civicrm_field['name']] = $field_definition_provider->getBaseFieldDefinition($civicrm_field);
       $fields[$civicrm_field['name']]->setRequired(isset($civicrm_required_fields[$civicrm_field['name']]));
     }
+    $fields['drupal_crud'] = BaseFieldDefinition::create('boolean')->setComputed(TRUE);
```
dsnopek, line 73: Question: Do we really need to declare a Drupal field for this? Or, could we declare just a plain class property on the `\Drupal\civicrm_entity\Entity\CivicrmEntity` class?

I don't know the answer to that question - it's a real question. :-) But it would seem nice to not have that be a Drupal field, but just a class property. Having it be a Drupal field means it could appear in some weird places, like grabbing the entity via the REST API.

kerasai: Just double-checked, without this the value doesn't persist throughout the save process and is not available in the Civi pre and post hooks.

I don't particularly like this either, open to other suggestions on how to make it work without defining as a field.

kerasai:
Added `try`/`finally` blocks around the save and delete calls and some other fixes.

Ready for another look.

src/Entity/CivicrmEntity.php
dsnopek: Ok! My feeling is if there isn't an easy answer, this is probably fine for now, and we can always fix later if someone comes up with a better way.

dsnopek:
:+1:

jackrabbithanna:
Great work, thanks guys :)
Merging #166

PR #167, opened 2018-08-23
update user create Rules action synchronizeUFMatch parameter logic, a…

jackrabbithanna:
…dd Display name as a option for the user name format


jackrabbithanna:
@aydun we had an issue come in and I need to make an adjustment to your update you made to the create user rules action code. 
You had contributed this commit:
https://github.com/eileenmcnaughton/civicrm_entity/commit/90973c9a0141fe76212e5d3b617bd546358221c2

Do you mind testing this new PR and see if it still solves your use case?

jackrabbithanna:
Merge happening!

PR #168, opened 2018-08-28
Use filter_xss in the entity property metadata info sanitize key

jackrabbithanna:
Although the API should provide the security, we can do something at the Drupal level for protecting from improper data entry from drupal forms and rules, and developers that use CiviCRM Entity as an API to build other Drupally things.


jackrabbithanna:
Merging #168

PR #170, opened 2018-08-30
CiviCRM timestamp fields are like Drupal datetime fields  …

dsnopek:
… not Drupal timestamp fields, which are UNIX time


jackrabbithanna:
Merging #170

PR #171, opened 2018-09-10
Convert datetime fields back from UTC for CiviCRM.

dsnopek:
This is a follow-up to #169 which fixed loading entities by converting the dates to UTC, but didn't fix saving the entities, which need to have the dates converted back from UTC before saving via the CiviCRM API. That's what this PR does!

I also snuck in some updates to comments to make the code related to this a little easier to understand, and removed an unused `elseif` branch (we don't use Timestamp fields anymore per #170).


jackrabbithanna:
Merging #171

PR #172, opened 2018-09-18
Fix logic for how to declare entity type when enabled or disabled

dsnopek:
This is a follow-up to #166

Using the latest on a live site, we discovered that it broke user registration. The error was:

```
Uncaught PHP Exception Drupal\Core\Entity\EntityStorageException: "The "civicrm_email" entity type does not exist." at /web/core/lib/Drupal/Core/Entity/Sql/SqlContentEntityStorage.php line 829
```

Basically, this came down to the logic for how we declare an entity type when enable or disabled didn't match how the new code decided that a Drupal entity type was available. The new logic is correct, it looks like the older code didn't match what I remember deciding!

Anyway, this PR fixes it in my testing

### DIFF
--- a/civicrm_entity.module
+++ b/civicrm_entity.module
@@ -55,41 +55,45 @@ function civicrm_entity_entity_type_build(array &$entity_types) {
       continue;
     }
 
+    $entity_type_info = [
+      'provider' => 'civicrm_entity',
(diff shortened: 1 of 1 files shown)


jackrabbithanna:
Merging #172

dsnopek:
Thanks!

PR #174, opened 2018-10-26
Fix for case where objectId not set in civicrm post hook

kerasai:
The `$objectId` variable is `NULL` for `ParticipantPayment` data in `hook_civicrm_post`. This causes issues when the value trickles down to Drupal's entity handling code and throws errors due to an array key being null.

This fix attempts to use the `::id` property from the object, if available.

I also deleted the `$entity` variable which was unused, as it's set later in the code.


jackrabbithanna:
Looks good, Thanks!
Merging #174

jackrabbithanna:
Pushed to D.O

PR #177, opened 2018-12-03
use API for CiviCRM groups options list for Assign/Unassign groups action

jackrabbithanna:
Before:
Query for option list would break on multi-lingual sites
After:
No error on creating action on multi-lingual sites.


jackrabbithanna:
Merging #177

Issue #179, opened 2019-05-16
check contact exist in group before deleting it  to avoid multiple log in watchdog


sunilpawar:
we have Drupal rule module configured to remove contact from group if certain criteria not meet.

Most of contact does not present in Group specified in Rules, This add N number of log watchdog, 

catch block only be called when there is error in removing contact from group.

jackrabbithanna:
Confirmed that if a contact is NOT in the group specified by the action settings, and the action runs a 'delete' action for the GroupContact entity, this causes the APIv3 to throw an exception, which is caught by a catch block and logs a message to the Drupal log. Thanks for the PR Sunil!

jackrabbithanna:
PR merged and pushed to Drupal.org, -dev branch, to be included in next release

PR #180, opened 2019-05-16
#179 check contact in group before calling delete action

sunilpawar:


jackrabbithanna:
Verified the error condition and that this PR improves the appropriate Rules action, merging #180

jackrabbithanna:
Pushed to -dev version on Drupal.org

PR #181, opened 2019-05-23
Bklaver d871

bklaver:
I don't actually want this to be merged right away, but thought I'd share my work.

I updated to Drupal 8.7.1 and received this message:

TypeError: Argument 1 passed to Drupal\civicrm_entity\Entity\Sql\CivicrmEntityStorageSchema::__construct() must implement
      interface Drupal\Core\Entity\EntityManagerInterface
, instance of Drupal\Core\Entity\EntityTypeManager given, called in htdocs/core/lib/Drupal/Core/Entity/Sql/SqlContentEntityStorage.php on line 292 in Drupal\civicrm_entity\Entity\Sql\CivicrmEntityStorageSchema->__construct() (line 46 of htdocs/modules/civicrm_entity/src/Entity/Sql/CivicrmEntityStorageSchema.php) #0 htdocs/core/lib/Drupal/Core/Entity/Sql/SqlContentEntityStorage.php(292): Drupal\civicrm_entity\Entity\Sql\CivicrmEntityStorageSchema->__construct(Object(Drupal\Core\Entity\EntityTypeManager), Object(Drupal\Core\Entity\ContentEntityType), Object(Drupal\civicrm_entity\CiviEntityStorage), Object(Drupal\Core\Database\Driver\mysql\Connection), Object(Drupal\Core\Entity\EntityManager))

I think this is because the constructor arguments were changed for the parent class here: https://git.drupalcode.org/project/drupal/blob/5d3322ebdab9c36ad1d06435c3ade0d8fa1e8f0a/core/lib/Drupal/Core/Entity/Sql/SqlContentEntityStorageSchema.php

I tried updating these classes to reflect the changes in their parents.  With very light testing, my site was still performing as before, but I'm pretty sure there's a problem I haven't found yet.  

I started this work, but have to hit pause.  Going to keep Drupal at 8.6.x for now.  Hopefully it helps someone.

### DIFF
--- a/src/CiviEntityStorage.php
+++ b/src/CiviEntityStorage.php
@@ -52,14 +52,18 @@ class CiviEntityStorage extends SqlContentEntityStorage implements DynamicallyFi
    *   The entity type definition.
    * @param \Drupal\Core\Database\Connection $database
    *   The database connection.
-   * @param \Drupal\Core\Entity\EntityManagerInterface $entity_manager
-   *   The entity manager.
+   * @param \Drupal\Core\Entity\EntityFieldManagerInterface $entity_field_manager
+   *   The entity field manager.
    * @param \Drupal\Core\Cache\CacheBackendInterface $cache
    *   The cache backend to be used.
    * @param \Drupal\Core\Language\LanguageManagerInterface $language_manager
    *   The language manager.
    * @param \Drupal\civicrm_entity\CiviCrmApiInterface $civicrm_api
    *   The CiviCRM API.
+   * @param \Drupal\Core\Entity\EntityTypeBundleInfoInterface $entity_type_bundle_info
+   *   The entity type bundle info.
+   * @param \Drupal\Core\Entity\EntityTypeManagerInterface $entity_type_manager
+   *   The entity type manager.
    */
   public function __construct(EntityTypeInterface $entity_type, Connection $database, EntityManagerInterface $entity_manager, CacheBackendInterface $cache, LanguageManagerInterface $language_manager, CiviCrmApiInterface $civicrm_api, ConfigFactoryInterface $config_factory) {
     parent::__construct($entity_type, $database, $entity_manager, $cache, $language_manager);
@@ -74,14 +78,15 @@ class CiviEntityStorage extends SqlContentEntityStorage implements DynamicallyFi
     return new static(
       $entity_type,
       $container->get('database'),
-      $container->get('entity.manager'),
+      $container->get('entity_field.manager'),
       $container->get('cache.entity'),
       $container->get('language_manager'),
-      $container->get('civicrm_entity.api'),
-      $container->get('config.factory')
+      $container->get('entity.memory_cache'),
+      $container->get('entity_type.bundle.info'),
+      $container->get('entity_type.manager')
     );
   }
-
+/
   /**
    * Initializes table name variables.
    */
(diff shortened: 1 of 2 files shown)


dsnopek:
Thanks! This is something that we noticed right away with 8.7 as well. The challenge is: how do we make this work for both Drupal 8.6 (which most people are still using) and 8.7 (which everyone will be moving to shortly)?

@mglaman Do you have any thoughts?

jackrabbithanna:
Since we never released a stable version, I'm fine with making it work for 8.7 and above only. We could peel away an "8.6-only" branch in this repo so people can have an option if necessary

src/CiviEntityStorage.php
```diff
@@ -74,14 +78,15 @@ public static function createInstance(ContainerInterface $container, EntityTypeI
     return new static(
       $entity_type,
       $container->get('database'),
-      $container->get('entity.manager'),
+      $container->get('entity_field.manager'),
```
mglaman, line 81: Why doesn't this explode if the construct call above still type hints `EntityManagerInterface`?

src/CiviEntityStorage.php
```diff
@@ -52,14 +52,18 @@ class CiviEntityStorage extends SqlContentEntityStorage implements DynamicallyFi
...
+   *   The entity field manager.
    * @param \Drupal\Core\Cache\CacheBackendInterface $cache
    *   The cache backend to be used.
    * @param \Drupal\Core\Language\LanguageManagerInterface $language_manager
    *   The language manager.
    * @param \Drupal\civicrm_entity\CiviCrmApiInterface $civicrm_api
    *   The CiviCRM API.
+   * @param \Drupal\Core\Entity\EntityTypeBundleInfoInterface $entity_type_bundle_info
+   *   The entity type bundle info.
+   * @param \Drupal\Core\Entity\EntityTypeManagerInterface $entity_type_manager
```
mglaman, line 65: I don't see these added to the parameters?

src/Entity/Sql/CivicrmEntityStorageSchema.php
```diff
@@ -81,12 +92,26 @@ public function requiresEntityStorageSchemaChanges(EntityTypeInterface $entity_t
     return FALSE;
   }
 
+
   /**
    * {@inheritdoc}
+   This will get rid of the "Mismatched entity and/or field definitions
+The following changes were detected in the entity type and field definitions." 
+Message on the status page. I don't think it should be left in though.
```
mglaman, line 100: But you can't modify the schema for these entities.

mglaman (commented):
Shoot. I think there might be an 8.6 and 8.7 BC layer. EntityManager implements all of the required interfaces. I'd like to give a second review of this code. I think it'd be fine to require 8.7, but there should be a BC layer possible.

bklaver:
Thanks for having a look @mglaman, I am out of my element here.

mglaman:
🤔 so here's one option. In `\Drupal\civicrm_entity\CiviEntityStorage` we stop injecting `civicrm_entity.api` and `config.factory` so that we do not override `createInstance` or `__construct`.

We can use `getCiviCrmApi()` and `getConfigFactory` wrappers. This mitigates a lot of the breaking changes.

And for `\Drupal\civicrm_entity\Entity\Sql\CivicrmEntityStorageSchema` .. I can't remember why the __construct was overridden. 

Experimenting with this.

mglaman:
It's late, so I couldn't test completely. @bklaver could you try #182? it removes some of the constructors and puts things in helpers to bridge 8.6 and 8.7 support.

@jackrabbithanna or @dsnopek could you test it on 8.6?

mglaman:
Should we close this for #182 ?

PR #182, opened 2019-05-25
8.6 and 8.7 support

mglaman:
An attempt to support 8.6 and 8.7 while working with internal structures

See #181 for original

### DIFF
--- a/src/CiviEntityStorage.php
+++ b/src/CiviEntityStorage.php
@@ -3,33 +3,20 @@
 namespace Drupal\civicrm_entity;
 
 use Drupal\civicrm_entity\Entity\CivicrmEntity;
-use Drupal\Core\Cache\CacheBackendInterface;
-use Drupal\Core\Config\ConfigFactoryInterface;
-use Drupal\Core\Database\Connection;
-use Drupal\Core\Database\DatabaseExceptionWrapper;
-use Drupal\Core\Database\SchemaException;
 use Drupal\Core\Entity\ContentEntityInterface;
-use Drupal\Core\Entity\ContentEntityStorageBase;
 use Drupal\Core\Entity\EntityInterface;
-use Drupal\Core\Entity\EntityManagerInterface;
-use Drupal\Core\Entity\EntityStorageException;
 use Drupal\Core\Entity\EntityTypeInterface;
-use Drupal\Core\Entity\Schema\DynamicallyFieldableEntityStorageSchemaInterface;
 use Drupal\Core\Entity\Sql\DefaultTableMapping;
 use Drupal\Core\Entity\Sql\SqlContentEntityStorage;
 use Drupal\Core\Field\FieldDefinitionInterface;
 use Drupal\Core\Field\FieldStorageDefinitionInterface;
 use Drupal\Core\Language\LanguageInterface;
-use Drupal\Core\Language\LanguageManagerInterface;
-use Drupal\Core\TypedData\Plugin\DataType\DateTimeIso8601;
-use Drupal\Core\TypedData\Plugin\DataType\Timestamp;
 use Drupal\field\FieldStorageConfigInterface;
-use Symfony\Component\DependencyInjection\ContainerInterface;
 
 /**
  * Defines entity class for external CiviCRM entities.
  */
-class CiviEntityStorage extends SqlContentEntityStorage implements DynamicallyFieldableEntityStorageSchemaInterface {
+class CiviEntityStorage extends SqlContentEntityStorage {
 
   /**
    * The CiviCRM API.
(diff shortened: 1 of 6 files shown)


src/CiviEntityStorage.php
```diff
@@ -3,33 +3,20 @@
...
-use Drupal\Core\Language\LanguageManagerInterface;
-use Drupal\Core\TypedData\Plugin\DataType\DateTimeIso8601;
-use Drupal\Core\TypedData\Plugin\DataType\Timestamp;
 use Drupal\field\FieldStorageConfigInterface;
-use Symfony\Component\DependencyInjection\ContainerInterface;
 
 /**
  * Defines entity class for external CiviCRM entities.
  */
-class CiviEntityStorage extends SqlContentEntityStorage implements DynamicallyFieldableEntityStorageSchemaInterface {
```
jackrabbithanna, line 32: Are we losing anything by NOT implementing DynamicallyFieldableEntityStorageSchemaInterface here?

src/CiviEntityStorage.php
```diff
@@ -46,40 +33,35 @@ class CiviEntityStorage extends SqlContentEntityStorage implements DynamicallyFi
...
-   *   The CiviCRM API.
+   * TODO: Inject after forced 8.7
    */
-  public function __construct(EntityTypeInterface $entity_type, Connection $database, EntityManagerInterface $entity_manager, CacheBackendInterface $cache, LanguageManagerInterface $language_manager, CiviCrmApiInterface $civicrm_api, ConfigFactoryInterface $config_factory) {
-    parent::__construct($entity_type, $database, $entity_manager, $cache, $language_manager);
-    $this->civicrmApi = $civicrm_api;
-    $this->configFactory = $config_factory;
+  private function getCiviCrmApi() {
+    if (!$this->civicrmApi) {
+      $this->civicrmApi = \Drupal::service('civicrm_entity.api');
```
dsnopek, line 43: Is there a way for us to use setter injection?

src/CiviEntityStorage.php
mglaman: Technically not, since it is implemented by SqlContentEntityStorage anyways

src/CiviEntityStorage.php
mglaman: I don't think so, like a `setCiviCrmAPI` method? I don't know when we would or could without worrying about more BC layers

dsnopek: Yeah, ok... If this was a normal service, we could do it in the services.yml, but entity storage is instantiated with it's own special magic. I guess we don't even care about injection at this point because we don't have any tests anyway. :-) We can convert this back to injection when we need it.

jackrabbithanna:
I'm a little hesitant to over complicate things or head off into the weeds to provide BC support that probably won't matter in a few months. I think as general goal we should have things as "standard" as possible, and continue to do so. We'd like to be able to keep up to date with Drupal APIs, follow standard practices, don't lose features just to maintain backward compatibility, and set ourselves up to have an easy transition to Drupal 9. For instance, I think it best to continue to use injection...and not compromise on that just for BC for a few months.
For me, our company is just about to start 3 D8 projects using this module, and all the sites will be on 8.7 or 8.8, so BC is not important at all to us.  We've been pretty hands off on this project so far but that should be changing the second half of the year. I wonder how we can gauge the importance of BC to people already using the module?  I can stir the pot on CiviCRM MM to find out. Would it be better to spin off a 4.x branch that works with Drupal 8.7 and above, and leave 3.x for BC? We could also make an alpha1 version that works for 8.6, and let the -dev release move forward with current APIs..knowing that at some point alpha2 would NOT work with 8.6
@dsnopek how important is BC to you? If it is not important, then should we just let people know they need to upgrade Drupal? If BC can be achieved while not compromising the long term vision of a stable D9 version, and getting overly exotic, maintaining current Drupal best practices, that's ok, but if not...

dsnopek:
Most modules try to support the two supported core minors (8.6 and 8.7 right now). Some do end up requiring the latest "early", but that's really a minority of modules.

Support for 8.6 is important for me right because we're not gonna update our customers to 8.7 for a few months, given the slew of update issues:

https://www.drupal.org/project/issues/search?issue_tags=8.7.0%20update

A bunch of them were fixed in 8.7.2, but a few still remain. Some of these include issues where the update process will break your database. :-( We're probably not the only ones holding off until all the update issues are fixed.

Matt's changes make sense to me. We're sub-classing a core storage class which is something we probably shouldn't have done because they could change it at any time, but we didn't have a choice due to some functionality that we needed being built into that class. There's open core issues to move that functionality out so that we don't have to do this. In any case, Matt's changes are basically removing some of the overrides we did to the methods from the parent class, so if the parent class changes them, then our overrides don't get in the way.

jackrabbithanna:
Ok good enough for me, avoid upgrade hell. Thanks for explaining that. I trust your and Matt's technical judgement. We can always move back to more standard approaches if necessary in 3-6 months when Drupal Core gets sorted out, or at least by the time a D9 version becomes necessary.

src/CiviEntityStorage.php
mglaman: :) We do have some tests

jackrabbithanna: Do you know what we need to do to get Drupal.org's infrastructure to run them automatically when we commit?

dsnopek: Ah, sorry! But I guess not unit tests if it doesn't matter if this is injected or not.

mglaman: I don't know how we could get DrupalCI to run, as it requires CiviCRM. We could on Travis. 

I was experimenting before for Webform CiviCRM https://github.com/mglaman/webform_civicrm/blob/travis-ci/.travis.yml

mglaman: @dsnopek correct, they're Kernel tests

mglaman:
@dsnopek @jackrabbithanna I rebased this to a single commit and did some cleanup. Drupal 8.7 and Drupal 8.8 will have better fixes for this. AKA: Drupal 9 will be easier. It's not nice and dependency injected, but it works and supports BC layers.

jackrabbithanna:
@dsnopek can you test on 8.6, while I test on 8.7 ? I don't have any 8.6 sites handy.

jackrabbithanna:
I did find a 8.6 site I could test on. 
Installs clean, available features look the same as before. I see the "manage" pages, the listing, Views works, Drupal "view" pages work..Editing contacts from Drupal forms works.

Same with 8.7
Appears to have feature equivalence, definitely with the broad strokes.

This looks good to me. 

This has been waiting on me for a long time, I'd like to push this out to D.O so that people using D8.7 can get going with it, who source this module from there or drupal/civicrm_entity composer installs.

Mark is finally getting around to CE again and Webform CiviCRM! I'm BAAAACKK.

Any reasons not to merge?

mglaman:
I see no reason not to

jackrabbithanna:
Merging #182 . Thanks @mglaman , you are a rock star! If nits pop up we'll deal with them as a new issue and a new PR.

dsnopek:
Sorry I wasn't around last week. Thanks, @mglaman & @jackrabbithanna for getting this merged!

jackrabbithanna:
Will make a new issue, getting reports and verified seeing Error messages on the Drupal Status page . Errors Found, "Entity/Field Definitions" ... seems to be 8.7 only ..
Confirmed reports ... running 'drush entity-updates' recognizes the condition, and runs seemingly without error, but doesn't fix the problem. I don't think it's caused by this PR, but something new to fix, related to D8.7

mglaman:
This is a general 8.7 WTF. Profile module has the same problems for it's
status entity key field

jackrabbithanna:
Hmm.. ok, functionality still seems to work. We'll put that on the list of WTFs :) Would you say I could tell people to not worry about it? We're getting some noise in the CiviCRM MM channels

mglaman:
Entity definition warnings have been a problem. I'd say don't worry about
it but maybe we need to look into it. I haven't been able to figure out the
Profile bug. But maybe there's something for this we can do

PR #183, opened 2019-07-25
getCount: allow optional params

mlutfy:
This makes the `getCount` params optional.

I initially encountered this while trying to uninstall the module. The `hasData` calls `getCount` on the various entities, without any params. Instead of providing an empty array from hasData, I think it makes sense to make the function argument optional.

### DIFF
--- a/src/CiviCrmApi.php
+++ b/src/CiviCrmApi.php
@@ -85,7 +85,7 @@ class CiviCrmApi implements CiviCrmApiInterface {
   /**
    * {@inheritdoc}
    */
-  public function getCount($entity, array $params) {
+  public function getCount($entity, array $params = []) {
     $this->initialize();
     $result = civicrm_api3($entity, 'getcount', $params);
     return is_int($result) ? $result : $result['result'];
--- a/src/CiviCrmApiInterface.php
+++ b/src/CiviCrmApiInterface.php
@@ -83,6 +83,6 @@ interface CiviCrmApiInterface {
    * @return int
    *   The number of entities.
    */
-  public function getCount($entity, array $params);
+  public function getCount($entity, array $params = []);
 
 }


mlutfy:
and unrelated, but I was just poking around, I had to patch `hasData` to do this:

```
  public function hasData() {
    if (substr($this->entityTypeId, 0, 8) == 'civicrm_') {
      return $this->civicrmApi->getCount(substr($this->entityTypeId, 8)) > 0;
    }

    return $this->civicrmApi->getCount($this->entityTypeId) > 0;
  }
```

otherwise it complained that "API (CivicrmActionSchedule, get) does not exist" (or various other entities).. but I'm not familiar enough with the code (and my patch probably breaks something else).

jackrabbithanna:
@mlutfy thanks the contrib .. I'll work this into my schedule to test

jackrabbithanna:
@mlutfy I don't see any problem with this patch in this PR, makes total sense, not seeing any negative consequences so far. I'm going to run our updates to hasData() for a little while on one of my builds and see how it goes. Since this method is returning a count / integer, it doesn't matter how the API call is made. It may be those API entities having errors, have entity types ids that are not quite matching the CiviCRM APIs is expecting, and that can be fixed too. Good reporting that maybe not complex to fix.

KarinG:
Running into this right now... 

I can drush pm:uninstall but I get a Fatal Error on: https://gfmam.org/admin/modules/uninstall

CiviCRM_API3_Exception: Unexpected result from getcountArray ( [error_code] => not-found [entity] => CivicrmActionSchedule [action] => get [is_error] => 1 [error_message] => API (CivicrmActionSchedule, get) does not exist (join the API team and implement it!) ) in civicrm_api3() (line 45 of /var/www/drupal/vendor/civicrm/civicrm-core/api/api.php).

Will test Mathieu's PR right now.

KarinG:
This PR works as described. I also get the "API (CivicrmActionSchedule, get) does not exist" - but after editing to:

```
  /**
   * {@inheritdoc}
   */
  public function hasData() {
    // KG before
    // return $this->getCiviCrmApi()->getCount($this->entityTypeId, []) > 0;
    // Mathieu's note on: https://github.com/eileenmcnaughton/civicrm_entity/pull/183
    if (substr($this->entityTypeId, 0, 8) == 'civicrm_') {
      return $this->civicrmApi->getCount(substr($this->entityTypeId, 8)) > 0;
    }
    return $this->civicrmApi->getCount($this->entityTypeId) > 0;
  }
```

I still can't get to Extend->Uninstall -> hitting: 

`Error: Call to a member function getCount() on null in Drupal\civicrm_entity\CiviEntityStorage->hasData() (line 265 of /var/www/drupal/web/modules/contrib/civicrm_entity/src/CiviEntityStorage.php) `

So right now with CiviCRM Entity installed a user can not get to the Extend->Uninstall

jackrabbithanna:
Would you like to try mlufty's patch in his second comment? If that works, we can make a PR for that too, and get that fixed. We've been using Drush to uninstall modules and have ignored this error for a while.

KarinG:
I did - and that does not work -> his   `public function hasData() {` edits produce:

`Error: Call to a member function getCount() on null in Drupal\civicrm_entity\CiviEntityStorage->hasData() 
`

And I don't know how to fix that.

KarinG:
I used this PR and then added on more to fix the Fatal Error on uninstall screen issue:
https://github.com/eileenmcnaughton/civicrm_entity/pull/184

KarinG:
@jackrabbithanna -> I did what you suggested and put in a more complete fix:

https://github.com/eileenmcnaughton/civicrm_entity/pull/184

dsnopek:
Thanks, Everyone! Closing this one in favor of https://github.com/eileenmcnaughton/civicrm_entity/pull/184

PR #184, opened 2019-09-12
Fix fatal error on modules/uninstall.

KarinG:
**Before:**
Can not uninstall modules via the GUI. 

**After:**
Can uninstall modules via the GUI.

**Steps to reproduce:**
Navigate to Extend -> Uninstall 
Observe Fatal Error

This PR replaces (and adds onto) https://github.com/eileenmcnaughton/civicrm_entity/pull/183

Note: I'm sure there is a better place to initialize 
`      $this->civicrmApi = $this->getCiviCrmApi();
`

But this works and admins can use the GUI again to uninstall modules

### DIFF
--- a/src/CiviCrmApiInterface.php
+++ b/src/CiviCrmApiInterface.php
@@ -83,6 +83,6 @@ interface CiviCrmApiInterface {
    * @return int
    *   The number of entities.
    */
-  public function getCount($entity, array $params);
+  public function getCount($entity, array $params = []);
 
 }
--- a/src/CiviEntityStorage.php
+++ b/src/CiviEntityStorage.php
@@ -258,7 +258,13 @@ class CiviEntityStorage extends SqlContentEntityStorage {
    * {@inheritdoc}
    */
   public function hasData() {
-    return $this->getCiviCrmApi()->getCount($this->entityTypeId, []) > 0;
+    if (!$this->civicrmApi) {
+      $this->civicrmApi = $this->getCiviCrmApi();
+    }
+    if (substr($this->entityTypeId, 0, 8) == 'civicrm_') {
+      return $this->civicrmApi->getCount(substr($this->entityTypeId, 8)) > 0;
+     }
+    return $this->civicrmApi->getCount($this->entityTypeId) > 0;
   }
   /**
    * {@inheritdoc}
--- a/src/CiviEntityStorage.php
+++ b/src/CiviEntityStorage.php
@@ -261,10 +261,7 @@ class CiviEntityStorage extends SqlContentEntityStorage {
     if (!$this->civicrmApi) {
       $this->civicrmApi = $this->getCiviCrmApi();
     }
-    if (substr($this->entityTypeId, 0, 8) == 'civicrm_') {
-      return $this->civicrmApi->getCount(substr($this->entityTypeId, 8)) > 0;
-     }
-    return $this->civicrmApi->getCount($this->entityTypeId) > 0;
+    return $this->civicrmApi->getCount($this->entityType->get('civicrm_entity')) > 0;
   }
   /**
    * {@inheritdoc}
--- a/src/CiviEntityStorage.php
+++ b/src/CiviEntityStorage.php
@@ -258,10 +258,7 @@ class CiviEntityStorage extends SqlContentEntityStorage {
    * {@inheritdoc}
    */
   public function hasData() {
-    if (!$this->civicrmApi) {
-      $this->civicrmApi = $this->getCiviCrmApi();
-    }
-    return $this->civicrmApi->getCount($this->entityType->get('civicrm_entity')) > 0;
+    return $this->getCiviCrmApi()->getCount($this->entityType->get('civicrm_entity')) > 0;
   }
   /**
    * {@inheritdoc}
--- a/src/CiviCrmApi.php
+++ b/src/CiviCrmApi.php
@@ -85,7 +85,7 @@ class CiviCrmApi implements CiviCrmApiInterface {
   /**
    * {@inheritdoc}
    */
-  public function getCount($entity, array $params) {
+  public function getCount($entity, array $params = []) {
     $this->initialize();
     $result = civicrm_api3($entity, 'getcount', $params);
     return is_int($result) ? $result : $result['result'];


mlutfy:
Thanks @karing, this definitely seems to work better than before. It's also catching that I have civicrm entities currently configured, and refusing to re-install, which I guess means that the code is doing what it's supposed to do.

That said, I have to admit I might not have a good understanding of the Entity module. Where can I see the entities, and how can I delete them? (this is on a dev site where the module was enabled, but never used, and there are no CiviCRM views)

KarinG:
I think check here: admin/structure/civicrm-entity
and then Settings -> will let you add more

mlutfy:
hmm, '/admin/structure/civicrm-entity' shows only a link to '/admin/structure/civicrm-entity/settings' and that page lists all available CiviCRM entities as checkboxes, but none of them are checked.

Capture d’écran de 2019-09-13 09-13-43

looking at the code, the `hasData` calls `getCount`, which checks for CiviCRM data in CiviCRM tables. Shouldn't it check against Drupal entities? It seems like a different logic than in the d7 module. (I understand that this code was there before your PR, and we can discuss in a separate issue if you prefer)

src/CiviCrmApiInterface.php
```diff
@@ -83,6 +83,6 @@ public function getOptions($entity, $field_name);
    * @return int
    *   The number of entities.
    */
-  public function getCount($entity, array $params);
+  public function getCount($entity, array $params = []);
```
dsnopek, line 86: This change (and in the implementation too) makes sense, we should do this.

src/CiviEntityStorage.php
```diff
@@ -258,7 +258,13 @@ public function countFieldData($storage_definition, $as_bool = FALSE) {
...
    */
   public function hasData() {
-    return $this->getCiviCrmApi()->getCount($this->entityTypeId, []) > 0;
+    if (!$this->civicrmApi) {
+      $this->civicrmApi = $this->getCiviCrmApi();
+    }
+    if (substr($this->entityTypeId, 0, 8) == 'civicrm_') {
+      return $this->civicrmApi->getCount(substr($this->entityTypeId, 8)) > 0;
+     }
+    return $this->civicrmApi->getCount($this->entityTypeId) > 0;
```
dsnopek: However, I think this change might be hiding a problem elsewhere, or maybe missing a utility function. Somewhere there's a list of civi -> drupal entity ids, or a utility to do that lookup, and that should be used here. Lemme poke around the code....

dsnopek: Alright, I've got it! This method should just be:

```
  public function hasData() {
    return $this->civicrmApi->getCount($this->entityType->get('civicrm_entity')) > 0;
  }
```

Basically, before it was trying to use the Drupal entity type id for the CiviCRM API, and this changes it to use the CiviCRM entity type id.

dsnopek:
> looking at the code, the hasData calls getCount, which checks for CiviCRM data in CiviCRM tables. Shouldn't it check against Drupal entities?

@mlutfy In Drupal 8, you're not required to have Drupal tables for an entity, you can have the entity be entirely based off a remote API, which is basically what civicrm_entity is doing.

However, in the case of uninstalling the module, this has a really weird effect, because you'll almost always have data in civicrm, but it should still be safe to uninstall civicrm_entity... That's another bug that'll need to be figured out!

src/CiviEntityStorage.php
KarinG: Thanks David! Ah ok - let me verify/make that edit

KarinG: Unfortunately that produces: `Error: Call to a member function getCount() on null in Drupal\civicrm_entity\CiviEntityStorage->hasData() (line 261 `

I did `drush cr` twice after making the edit

I still need to do this (in order to not get the fatal error on the uninstall screen):
```

    if (!$this->civicrmApi) {
      $this->civicrmApi = $this->getCiviCrmApi();
    }
    return $this->civicrmApi->getCount($this->entityType->get('civicrm_entity')) > 0;


```

KarinG:
What happens on Uninstall is that you can’t uninstall civicrm_entity module b/c there are entities -> and the user is prompted to delete those one by one (first).

KarinG:
@dsnopek - I've posted screenshots of the Uninstall process on: https://github.com/eileenmcnaughton/civicrm_entity/issues/185
It's like Hotel California :-) you can (try) check out, but you can never leave!

@mlutfy - on /admin/structure/civicrm-entity/settings you can select which one(s) you want:
for example:
image

Save that and then these will appear on /admin/structure/civicrm-entity
image

and if I hit e.g. CiviCRM Contact Entities -> I see all my Contacts -> in this small d8 CiviCRM:
image

Hit Edit -> you get all fields for that Contact ->  adding a value and saving it -> works. I am first and foremost interested in the View Integration and with the new integration you can now do Entities instead of Fields -> like:

image

Change the display to e.g. Bootstrap Grid or Bootstrap Cards -> and you've got a business card like display within just a few clicks

jackrabbithanna:
You can "disable" the module like this, in the meantime, until a solution is arrived at:
drush eval "$module_data = \Drupal::config('core.extension')->get('module'); unset($module_data['civicrm_entity'']); \Drupal::configFactory()->getEditable('core.extension')->set('module', $module_data)->save();"

src/CiviEntityStorage.php
dsnopek: Hm. I'm not sure how this bit helps:

```
if (!$this->civicrmApi) {
  $this->civicrmApi = $this->getCiviCrmApi();
}
```

There is no `$this->getCiviCrmApi()` method on `CiviEntityStorage` AND `$this->civicrmApi` should never be NULL because it's required by the class constructor...

dsnopek: I just tested without that bit of code and the "Uninstall" page is loading for me... Not sure what we're doing differently in our environments?

KarinG: Wicked! I've seen this on two sites/two servers thus far. Let me try again (going backwards on d8.semper-it.com):

**1. I can get to Uninstall screen:**
image

**2. Commenting out these lines:**
image

**3. Reload Uninstall Screen:**
`Error: Call to a member function getCount() on null in Drupal\civicrm_entity\CiviEntityStorage->hasData() (line 264 of /var/www/drupal/web/modules/contrib/civicrm_entity/src/CiviEntityStorage.php) #0 /var/www/drupal/web/core/lib/Drupal/Core/Entity/ContentUninstallValidator.php(56): Drupal\civicrm_entity\CiviEntityStorage->hasData() `

If you can think of anythings else I can try - let me know.

KarinG: Ok mystery solved -> `$this->getCiviCrmApi` got introduced in https://github.com/eileenmcnaughton/civicrm_entity/pull/182 - which David does not use yet.

jonhalle:
Sorry to ask in the wrong place but I don't understand KarinG's last screenshot and I think it is at the root of the problems I am having with Civicrm Entity which have been frustrating me for months. If I want to set up a View using Civicrm Entities provided by Civicrm Entity, how do I get to the point shown in that screenshot where you see "Show: Civicrm Address" and it's not being referenced as a field? Views|Add View... then what?

src/CiviEntityStorage.php
```diff
@@ -258,7 +258,10 @@ public function countFieldData($storage_definition, $as_bool = FALSE) {
    * {@inheritdoc}
    */
   public function hasData() {
-    return $this->getCiviCrmApi()->getCount($this->entityTypeId, []) > 0;
+    if (!$this->civicrmApi) {
+      $this->civicrmApi = $this->getCiviCrmApi();
+    }
+    return $this->civicrmApi->getCount($this->entityType->get('civicrm_entity')) > 0;
```
dsnopek: @KarinG Could you change this function to just:

```
return $this->getCiviCrmApi()->getCount($this->entityType->get('civicrm_entity')) > 0;
```

The `getCiviCrmApi()` already does the bit with caching the service in the class property, so there's no need to duplicate that.

But after that, I think this is ready to merge! The issue with uninstall prompting you to delete all your Civi entities is a separate issue (which should get solved nonetheless, but we don't need to in this PR :-))

KarinG: Hi @dsnopek - as we discussed over a call - in D8.7 -> I need to: https://github.com/eileenmcnaughton/civicrm_entity/pull/184#discussion_r324813678 (items #2 and #3).

KarinG: I don't know how to solve that without simply checking 

`if (!$this->civicrmApi) {`

dsnopek: @KarinG I'm not sure I understand. If the whole function is:

```
  public function hasData() {
    return $this->getCiviCrmApi()->getCount($this->entityType->get('civicrm_entity')) > 0;
  }
```

You shouldn't need to check `if (!$this->civicrmApi) {` because that's taken care of in `$this->getCiviCrmApi()`, there's no reason to duplicate that check here.

Or am I missing something?

KarinG: I don't understand this nearly as well as you do - but if I comment out the check:

```
  public function hasData() {
    //if (!$this->civicrmApi) {
    //  $this->civicrmApi = $this->getCiviCrmApi();
    ///}
    return $this->civicrmApi->getCount($this->entityType->get('civicrm_entity')) > 0;
  }

```

Then 
`root@9f03c6ef5d47:/var/www/drupal/web/modules/contrib/civicrm_entity/src# drush cr`

**And hit:**
`/admin/modules/uninstall`

I get many (exactly one for every CiviCRM Entity under: `/admin/structure/civicrm-entity`)
`Error: Call to a member function getCount() on null in Drupal\civicrm_entity\CiviEntityStorage->hasData() `

**Our env:**
Drupal 8.7.5
php -version: PHP 7.3.7 (cli) (built: Jul 17 2019 22:29:15) ( NTS )

KarinG: Hold on -> I see you have an additional () -> let me add that.

KarinG: Ok phew! Makes sense now. Sorry missed that first time around! Going to change the PR now.

dsnopek: Thanks! Merging :-)

PR #186, opened 2019-09-30
Fix for not empty time on end date field

mfrausto-mdw:
**Before:**
Got an error when try to save an event when end date partially filled

**After:**
Get a message specifying you need to fill both date and time in order to save the event

**Steps to reproduce:**
Admin -> Content -> Add Content -> Event
Fill all required fields including start date
Fill only date and leave time empty on end date.
Try to save the event.

**About the change:**
When end date is partially filled is an array,  but if field is totally filled is a string


dsnopek:
Thanks, this fixes the problem in my testing!

PR #187, opened 2019-10-01
Fix triggering Drupal hooks when entity updated in Civi

dsnopek:
I'm not entirely sure when/how this got broken, but it might be from a Drupal core change. In any case, Drupal hooks are no longer being fired when an entity is updated via CiviCRM. We noticed this primarily with caching: a user would update a Civi Event in the Civi admin, but the Drupal display page wouldn't change.

It looks like the problem comes from the magic that Drupal adds to Entity objects to access Drupal fields as properties. We were making 'drupal_crud' be a computed Drupal field, but then checking it via `!empty($entity->drupal_crud)` which is now always evaluating as TRUE.

The change in this PR is making it a normal public property, and not a computed Drupal field, and then everything goes back to working as it should.

I tested this by CRUD'ing (creating, reading, updating, deleting) a Civi Event from both the Civi admin and Drupal, and everything seems to be working.

### DIFF
--- a/src/Entity/CivicrmEntity.php
+++ b/src/Entity/CivicrmEntity.php
@@ -23,6 +23,17 @@ use Symfony\Component\Validator\ConstraintViolation;
 class CivicrmEntity extends ContentEntityBase {
 
   /**
+   * Flag to denote if the entity is currently going through Drupal CRUD hooks.
+   *
+   * We need to trigger the Drupal CRUD hooks when entities are edited in Civi,
+   * but we need a way to ensure they aren't double triggered when already
+   * going through the Drupal CRUD process.
+   *
+   * @var bool
+   */
+  public $drupal_crud = FALSE;
+
+  /**
    * {@inheritdoc}
    */
   public function save() {
@@ -70,7 +81,6 @@ class CivicrmEntity extends ContentEntityBase {
       $fields[$civicrm_field['name']] = $field_definition_provider->getBaseFieldDefinition($civicrm_field);
       $fields[$civicrm_field['name']]->setRequired(isset($civicrm_required_fields[$civicrm_field['name']]));
     }
-    $fields['drupal_crud'] = BaseFieldDefinition::create('boolean')->setComputed(TRUE);
     return $fields;
   }
 


KarinG:
I just applied this to my D8WFC test instance where I have the webform create/store an address and then I retrieve it with a CiviCRM Entity View. I initially struggled with caching; then I did a number of things to try help with that (disabled Views Caching; set Cache: none for the View itself, set browser and proxy max-age to: no caching & now added this PR. I can now consistently submit the webform as anonymous -> retrieve the results.

dsnopek:
@KarinG Are you saying that with just this PR it's enough to fix your caching issues? Or, only together with all the other configuration changes too? Thanks!

KarinG:
@dsnopek - I'm not sure. Let me try disable all but this PR:

**rollback:**
Within View -> Advanced -> Caching: None -> change to Tag based
/admin/structure/views/settings/advanced -> deselect Disable views caching

**keep:**
/admin/config/development/performance -> browser and proxy cache maximum age: <no caching>

Test (as anonymous user):
https://d8.semper-it.com/node/41 [a D8.7]

And that works! I get my newly entered address displayed on the Page view within seconds after submitting. 

Triple checking that I have your PR still in my code base. Yes I do (see below). 

Let's merge this in! I can say it definitively works a lot better then it did before (when I had to also disable all Views caching to get 'instant' playback of CiviCRM data submitted via D8WFC).

```
diff --git a/src/Entity/CivicrmEntity.php b/src/Entity/CivicrmEntity.php
index b632e53..9734570 100644
--- a/src/Entity/CivicrmEntity.php
+++ b/src/Entity/CivicrmEntity.php
@@ -23,6 +23,17 @@ use Symfony\Component\Validator\ConstraintViolation;
 
   /**
+   * Flag to denote if the entity is currently going through Drupal CRUD hooks.
+   *
+   * We need to trigger the Drupal CRUD hooks when entities are edited in Civi,
+   * but we need a way to ensure they aren't double triggered when already
+   * going through the Drupal CRUD process.
+   *
+   * @var bool
+   */
+  public $drupal_crud = FALSE;
+
+  /**
    * {@inheritdoc}
    */
   public function save() {
@@ -70,7 +81,6 @@ class CivicrmEntity extends ContentEntityBase {
       $fields[$civicrm_field['name']] = $field_definition_provider->getBaseFieldDefinition($civicrm_field);
       $fields[$civicrm_field['name']]->setRequired(isset($civicrm_required_fields[$civicrm_field['name']]));
     }
-    $fields['drupal_crud'] = BaseFieldDefinition::create('boolean')->setComputed(TRUE);
     return $fields;
   }

```

dsnopek:
Awesome, thanks! Merging :-)

PR #188, opened 2019-10-15
Ensure foreign key integers are treated as entity references

mglaman:
This ensures `\CRM_Utils_Type::T_INT` are properly treated as foreign keys and turn into entity references. I _think_ this will resolve (for the most part) Views relationships.

I have not tested beyond updating the unit test.


KarinG:
Testing!

I pulled in this PR - and can't drush cr - I'm getting a PHP error:

```
root@9f03c6ef5d47:/var/www/drupal/web/modules/contrib/civicrm_entity# drush cr
In EntityTypeManager.php line 150:                                            
  The "civicrm_msg_template" entity type does not exist.  
```
                                                          
That table does exist: `civicrm_msg_template` - but really isn't all that useful to construct views/output. Is there some selection of tables that become entities?

mglaman:
Ah, the code should make sure the referenced table is a supported CiviCRM entity.

mglaman:
@KarinG my machine is borked today. I updated the check and the test passed -- a non-entity foreign key stayed as an integer and not an entity reference.

I was not able to manually test, due to my Docker exploding.

KarinG:
It is working!!!

**Address Directory - ici:**
image

PS - I really like the checkmarks for 1/0 fields - that's a cool D8 Views Output Format. 

**Relationship - voila:**
image

KarinG:
@jackrabbithanna or @dsnopek - this can be merged!

MegaphoneJon:
I did some testing and it seems like the entity references are limited to a single direction.  You can create a Views relationship from Address to Contact, but not Contact to Address.  I couldn't find a way to satisfy my use case of "Create a directory of active members with a link to their website", for instance.

API4 handles this with the `SchemaMap` object, which the `SchemaMapBuilder` object builds.  It handles [FK backreferences](https://github.com/civicrm/org.civicrm.api4/blob/master/Civi/Api4/Service/Schema/SchemaMapBuilder.php#L126).  It also handles other oddities gracefully (custom field tables, many-to-many joins).

I wonder if the solution is to make API4 a dependency of D8CE and use the API4 methods?  It's going to be in core imminently.

mglaman:
Drupal should automatically provide the reverse relationship. The entity reference field views data handler should do this. Worth reviewing there.

MegaphoneJon:
@mglaman Interesting.  Here are screenshots of two fresh Views (using D8CE with this PR applied).  In addition to my last observation:
* Not all the FKs on `civicrm_address` are present.  I.e. the FKs to `civicrm_state_province` and `civicrm_county` are missing.
* The first relationship on Adresses is to itself.  I presume that's because of the FK on `civicrm_address.master_id`.  However, the UI doesn't indicate this.

Finally - are you seeing what I'm seeing here?  I don't want to waste either of our times if this is down to a misconfiguration on my end.

Selection_689
Selection_690

mglaman:
I'm not at a computer to test right now. But part of the problem might be how the foreign keys are represented period for those are foreign keys integer columns? 

It currently only assumes that all fields which have T_INT and a foreign key class are remote references. The proper fix with how we did check if there is a foreign key class first and then use that to reference the other entity.

For instance I'm not able to check if addresses and state provinces use integer based identifiers

dsnopek:
Since this PR does legitimately fix some relationships, should we merge this and make a new issue/PR to address the remaining ones? Or would you folks rather hash that out here?

KarinG:
I vote for merge.

mglaman:
I vote merge and open an issue for reverse and more specific missing relationships.

dsnopek:
Merged! Thanks, Everyone. Looking forward to continuing to get more references working. :-)

KarinG:
@MegaphoneJon - Thanks Jon for looking at this. I'll open an issue and will reference your notes.

PR #190, opened 2019-10-19
Take into consideration the date definition settings

### LINKED ISSUE #189: Date of Birth Field -> renders blank when creating a View
**Steps to reproduce:**
1. (top part): View of CiviCRM Contact -> fields: Name, Birth Date, Created Date, Contact id.
2. (middle part): Birth Date is blank -> Created Date is ok.
3. (bottom part): in CiviCRM Contact with id = 2 -> definitively has a Date of Birth -> June 5, 1970

Same issue when using Contact Rendered Entity.

image

mglaman:
Fixes #189.

### DIFF
--- a/src/CiviEntityStorage.php
+++ b/src/CiviEntityStorage.php
@@ -11,6 +11,8 @@ use Drupal\Core\Entity\Sql\SqlContentEntityStorage;
 use Drupal\Core\Field\FieldDefinitionInterface;
 use Drupal\Core\Field\FieldStorageDefinitionInterface;
 use Drupal\Core\Language\LanguageInterface;
+use Drupal\datetime\Plugin\Field\FieldType\DateTimeItem;
+use Drupal\datetime\Plugin\Field\FieldType\DateTimeItemInterface;
 use Drupal\field\FieldStorageConfigInterface;
 
 /**
@@ -164,6 +166,7 @@ class CiviEntityStorage extends SqlContentEntityStorage {
 
     // get all the fields
     $fields = $this->getCiviCrmApi()->getFields($this->entityType->get('civicrm_entity'));
+    $field_names = [];
     foreach ($fields as $field) {
       $field_names[] = $field['name'];
     }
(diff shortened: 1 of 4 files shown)


KarinG:
Roger that! I now see a date of birth:

image

@jackrabbithanna or @dsnopek - this one can be merged too!

src/CiviEntityStorage.php
```diff
@@ -362,13 +365,15 @@ protected function initFieldValues(ContentEntityInterface $entity, array $values
...
-            $item_values[$delta][$main_property_name] = (new \DateTime())->setTimestamp($item[$main_property_name])->format(DATETIME_DATETIME_STORAGE_FORMAT);
+            $item_values[$delta][$main_property_name] = (new \DateTime('now', new \DateTimeZone('UTC')))->setTimestamp($item[$main_property_name])->format(DateTimeItemInterface::DATETIME_STORAGE_FORMAT);
           }
           else {
+            $datetime_format = $definition->getSetting('datetime_type') === DateTimeItem::DATETIME_TYPE_DATE ? DateTimeItemInterface::DATE_STORAGE_FORMAT : DateTimeItemInterface::DATETIME_STORAGE_FORMAT;
             // CiviCRM gives us the datetime in the users timezone (or no
             // timezone at all) but Drupal expects it in UTC. So, we need to
             // convert from the users timezone into UTC.
-            $item_values[$delta][$main_property_name] = (new \DateTime($item[$main_property_name], new \DateTimeZone(drupal_get_user_timezone())))->setTimezone(new \DateTimeZone('UTC'))->format(DATETIME_DATETIME_STORAGE_FORMAT);
+            $datetime_value = (new \DateTime($item[$main_property_name], new \DateTimeZone('UTC')))->format($datetime_format);
```
dsnopek: I haven't tested this, but I suspect the changes here to UTC are not quite right. CiviCRM doesn't keep the datetimes in UTC time, it keeps them in local time. So, a lot of the unusual date formulations here are converting the time in the database from local time into UTC time, because Drupal expects them to be in UTC.

@mglaman Have you tested this with other timezones?

mglaman: I have not. But I realized there was proof of something weird in the test assertions.

https://github.com/eileenmcnaughton/civicrm_entity/pull/190/files#diff-22cf275b4dbef73100eb9fe7e1347fb8L44

```
-    $this->assertEquals('2018-05-02T07:00:00', $entity->get('start_date')->value);
+    $this->assertEquals('2018-05-02T17:00:00', $entity->get('start_date')->value);
```

The test data has the value 

```
        'start_date' => '2018-05-02 17:00:00',
        'event_start_date' => '2018-05-02 17:00:00',
        'end_date' => '2018-05-04 17:00:00',
        'event_end_date' => '2018-05-04 17:00:00',
```

The original test was showing a 10-hour difference.  With the adjustment, it's now correct. Drupal also executes it's PHPUnit with the timezone set to Sydney, Australia

```php
// Set the default timezone. While this doesn't cause any tests to fail, PHP
// complains if 'date.timezone' is not set in php.ini. The Australia/Sydney
// timezone is chosen so all tests are run using an edge case scenario (UTC+10
// and DST). This choice is made to prevent timezone related regressions and
// reduce the fragility of the testing system in general.
date_default_timezone_set('Australia/Sydney');
```

It's also more concrete than trusting `drupal_get_user_timezone` which is per-user. Maybe we need to check the timezone setting in CiviCRM instead? Because that's more reliable than Drupal's configured timezone, technically.

dsnopek: The reason I think it's using the user timezone, is that we want the timezone to _appear_ to be unprocessed. So, if an event has '2020-01-01T10:00:00' in the database, we want Drupal to show it to the user as exactly January 1st, 2020 @ 10am, no matter what the user has their timezone set at.

So, that mess of original code is basically saying, "Let's interpret the database datetime as happening in the current user's timezone, and then turn it into UTC for Drupal, which will later turn it BACK into the user's timezone, so it's displayed exactly as it is in the database." Unfortunately, while super weird, I think this is actually correct. We should have some tests for this.

KarinG: I ran into a timezone issue with D8/CiviCRM myself recently. Just D8/CiviCRM native -> Contributions entered are coming in (have received dates) anywhere between user time zone (from D8) and UTC: https://civicrm.stackexchange.com/questions/33541/timezone-issue-authenticated-vs-anonymous-a-6h-time-warp

mglaman:
@dsnopek the timezone changes have been reverted, should keep the same timezone handling as they currently exist in 8.x-3.x.

dsnopek:
@mglaman Looks like there's some conflicts to resolve (probably from merging that other PR). Also, assuming it wouldn't be a huge like, could you add a test that verifies the timezone wackiness in here? I'm worried a change could slip past that removes later when we forget why we're doing this :-)

mglaman:
@dsnopek I can see about hardening the timezone logic.

KarinG:
@mglaman - I ran into a timezone issue with the recording of Contributions - and put up this PR for the D8 integration with CiviCRM https://github.com/civicrm/civicrm-core/pull/15794

mglaman:
@dsnopek for the delay! added three tests to confirm the timezone changes.

dsnopek:
Awesome, thanks! This looks good to me :-)

PR #192, opened 2019-10-27
Add county and state_province to improve address book use case

### LINKED ISSUE #191: Not all the FKs on civicrm_address are present. I.e. the FKs to civicrm_state_province and civicrm_county are missing.
Reported by Jon in: https://github.com/eileenmcnaughton/civicrm_entity/pull/188

mglaman:
Fixes #191 

The problem wasn't the code but that the entity definitions were not exposed to Drupal. Also fixed the failing tests due to missing DAO classes (civicrm isn't autoloaded)


dsnopek:
I wonder if we should do views support and normal entity support separately? I can't imagine a situation where you'd want to expose counties and states/provinces as normal entities... What did the D7 version of civicrm_entity do?

mglaman:
> What did the D7 version of civicrm_entity do?

No idea, maybe that's where the civicrm_views module filled a gap?

> I wonder if we should do views support and normal entity support separately? 

I agree. my thought was "we'll mirror Drupal entity definitions from CiviCRM, but only certain ones can be managed through Drupal". So county + state_province entity definitions are available _but_ you cannot toggle them in the UI.

KarinG:
Hi @mglaman - 

county and state_province use pseudoconstants - were actually fine; but with this PR they are missing (see screenshot below - compare that to my screenshot in https://github.com/eileenmcnaughton/civicrm_entity/issues/191 . Jon also acknowledged he had those working. Sorry - should have closed the issue and opened a new one!

image

I think the real issue is that if you start with 
civicrm_address as base -> you can create a relationship to civicrm_contact;

but if you start with
civicrm_contact as base -> you can not create a relationship to civicrm address; 

So we've got one way relationships and making a Membership directory that spans three tables -> you would need to start with civicrm_contact; then relationship to civicrm_address + relationship to civicrm_member -> that we can't do yet.

mglaman:
_oh_ okay. I'll close this and rebuild based on 

> civicrm_contact as base -> you can not create a relationship to civicrm address;

jackrabbithanna:
I actually like this PR, I'd like to see state/province and county integrated as entity types. 
There are times when you want state/province in Views, and @dsnopek make a patch that fixed an issue with the Country entity type in Views, that would apply to this.
I'm inclined to merge it unless there are major objections.

jackrabbithanna:
@mglaman if you feel like fixing the merge conflict I'd merge this.

mglaman:
@jackrabbithanna I'm not available to make any headway on this anymore, unfortunately.

jackrabbithanna:
Ok no worries, thanks for all you have done! I'm closing this PR for now, and when we need it, we'll add these two entity types.

PR #197, opened 2019-11-21
List integer filters

### LINKED ISSUE #196: All fields are treated as regular text fields; option lists don't populate
To illustrate the issue, consider these two screenshots from D7/D8:

**D7**:
Selection_721

**D8:**
Selection_720

The D7 Views knows that this is an option list and not a text field.  It also displays the label and not the primary key from `civicrm_state_province`.

In addition to not displaying as a select box, an exposed filter doesn't work unless you enter the key of the value - e.g. searching on country, you must enter `1228` instead of `United States`.

Knowledge about data types extends to other areas as well - e.g. I can pull the Image URL from `civicrm_contact`, but it displays as a plaintext URL, it's not recognized as an image data type, so e.g. I can't use image styles on it.

mglaman:
Fixes #196

### DIFF
--- a/src/CivicrmEntityViewsData.php
+++ b/src/CivicrmEntityViewsData.php
@@ -125,10 +125,10 @@ class CivicrmEntityViewsData extends EntityViewsData {
               '@entity' => $this->entityType->getLabel(),
             ];
             $data[$target_base_table][$pseudo_field_name]['relationship'] = [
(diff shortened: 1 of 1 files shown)


dsnopek:
Thanks!

Code looks good and fixed the problem in my testing.

PR #199, opened 2019-12-16
Fixes #195 CiviCRM Entity breaks in Drupal 8.8 - removes constructor,…

### LINKED ISSUE #195: CiviCRM Entity breaks in Drupal 8.8
Installing the module fails with errors:

> PHP Fatal error:  Uncaught Error: Call to a member function getFieldStorageDefinitions() on null in /var/www/html/distro8.skvare.com/web/core/modules/views/src/EntityViewsData.php:132
> Stack trace:
> #0 /var/www/html/distro8.skvare.com/web/core/modules/views/src/EntityViewsData.php(415): Drupal\views\EntityViewsData->getFieldStorageDefinitions()
> #1 /var/www/html/distro8.skvare.com/web/modules/custom/civicrm_entity/src/CivicrmEntityViewsData.php(117): Drupal\views\EntityViewsData->mapFieldDefinition('civicrm_action_...', 'id', Object(Drupal\Core\Field\BaseFieldDefinition), Object(Drupal\Core\Entity\Sql\DefaultTableMapping), Array)
> #2 /var/www/html/distro8.skvare.com/web/core/modules/views/views.views.inc(178): Drupal\civicrm_entity\CivicrmEntityViewsData->getViewsData()
> #3 [internal function]: views_views_data()
> #4 /var/www/html/distro8.skvare.com/web/core/lib/Drupal/Core/Extension/ModuleHandler.php(392): call_user_func_array('views_views_dat...', Array)
> #5 /var/www/html/distro8.skvare.com/web/core/modules/views/src/ViewsData.php( in /var/www/html/distro8.skvare.com/web/core/modules/views/src/EntityViewsData.php on line 132
>  [warning] Drush command terminated abnormally. Check for an exit() in your Drupal site

stesi561:
… and createInstance as code was duplicated from parent, and had been updated in a breaking way in the parent.

As well as removes a few not unused "use ..." statements.

I'm not sure why this code was duplicated in the first place. The only reason I can see is that the constructor did enforce a more specific subclass on the Data store - CiviEntityStorage - However I can't get the code to run easily by keeping this and it does appear to work without it.

Additionally I've removed a heap of "use" statements that appear to only be used for the constructor.

### DIFF
--- a/src/CivicrmEntityViewsData.php
+++ b/src/CivicrmEntityViewsData.php
@@ -2,55 +2,15 @@
 
 namespace Drupal\civicrm_entity;
 
-use Drupal\Core\Entity\EntityManagerInterface;
 use Drupal\Core\Entity\EntityTypeInterface;
-use Drupal\Core\Extension\ModuleHandlerInterface;
+
 use Drupal\Core\Field\FieldDefinitionInterface;
 use Drupal\Core\StringTranslation\StringTranslationTrait;
-use Drupal\Core\StringTranslation\TranslationInterface;
 use Drupal\views\EntityViewsData;
-use Symfony\Component\DependencyInjection\ContainerInterface;
 
 class CivicrmEntityViewsData extends EntityViewsData {
 
   use StringTranslationTrait;
-
-  /**
-   * Constructs an EntityViewsData object.
-   *
-   * @param \Drupal\Core\Entity\EntityTypeInterface $entity_type
-   *   The entity type to provide views integration for.
-   * @param \Drupal\civicrm_entity\CiviEntityStorage $storage_controller
-   *   The storage handler used for this entity type.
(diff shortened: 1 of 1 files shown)


stesi561:
I've only tested a basic view. And I think this needs a review by someone more familiar with what's going on here that me, or at least some more thorough testing but wanted to get this up as might help anyone who is stuck on 8.8.0 with a broken civicrm entity.

jackrabbithanna:
Thanks for the contribution! This looks like a reasonable PR. Now __construct() would be handled by the EntityViewData class instead of CivicrmEntityViewsData .. this may be the way to go for backwards compatibility with 8.6/8.7 as well. I'll do some testing.

mglaman:
This looks good to me. I think it was left around solely for 8.6 support, which is no longer supported.

jackrabbithanna:
Merging #199

PR #203, opened 2020-02-24
Readme & Version constraint fixes.

homotechsual:
Two changes here for composer.json:

1. Changes the composer.json file to remove the repository information - this isn't referenced except in the "root/project" composer.json file. 
2. Adds appropriate version constraints to composer.json to enable composer based installation when using `civicrm/civicrm-drupal-8` with a version (which you really should be doing - using dev-master is dangerous!)

Adds a basic readme with composer installation steps.

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,4 +22,4 @@ For example:
        "civicrm/civicrm-drupal-8": "^5.19",
 ```
 
-By default the **RoundEarth** method uses `dev-master` for `civicrm/civicrm-drupal-8` - this is dangerous and not recommended!
+By default the **RoundEarth** method uses `dev-master` for `civicrm/civicrm-drupal-8` - this is [dangerous and not recommended](https://lab.civicrm.org/dev/drupal/issues/87#note_23534)!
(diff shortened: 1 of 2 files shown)


jackrabbithanna:
We talked about this a bit in civi mattermost chat, I was thinking about which version of CiviCRM was the first to work reliably as it does now with D8 as what the constraint should be in the CE composer.json

CiviCRM Entity for D8 will work with any version of CiviCRM, that works with Drupal 8. At it's bottom it's all just CiviCRM APIv3.

I think ~5.0 would be ok in this case for the version constraint in the composer.json. It will pull the latest version in the 5.x series, so current versions would be totally fine.

homotechsual:
I've tested this install method on 19 sites (okay technically our CD runner tested it :-p) and it works (we used a forked repo with the version constraints changed!)

The one thing I think I should note in the readme is that to use this you should change your composer.json file to specify the version for `civicrm-drupal-8` - thoughts?

jackrabbithanna:
I've been ONLY requiring civicrm-drupal-8 in our custom composer template, but using a forked version of the civicrm-drupal-8 repo we maintain. Then in that forked civicrm-drupal-8 repo, specifying civicrm-core as a dependency of that. The main reason to do that was to have different branches / tags with the right version dependency.   So for example I'd have a tag 5.13.8 of civicrm---drupal-8 repo, which would in it's composer.json require civicrm-core 5.13.8. 

So specifying version for civicrm-drupal-8 makes sense.

homotechsual:
Fixed the readme!

jackrabbithanna:
Merging #203

Issue #200, opened 2019-12-19
Custom fields can't be used in filters - Port D7 Core Custom Field Integration to CiviCRM Entity


jamie-tillman:
To reproduce, create a custom field for Contacts (mine was a multi-valued select list)
Create a view for Contact Entities and add the custom field to the filter (I exposed mine to the user with the default "Is one of" operator)
Execute the view with no filter value selected and it works fine.
Select a filter value and Apply it and you get an error like the following:
```
Drupal\Core\Database\DatabaseExceptionWrapper: Exception in Member Directory[member_directory]: SQLSTATE[42S22]: Column not found: 1054 Unknown column 'civicrm_contact.custom_2' in 'where clause': SELECT civicrm_contact.id AS id, contact_id_civicrm_contact.id AS contact_id_civicrm_contact_id FROM {civicrm_contact} civicrm_contact INNER JOIN {civicrm_membership} contact_id_civicrm_contact ON civicrm_contact.id = contact_id_civicrm_contact.contact_id WHERE ((contact_id_civicrm_contact.status_id IN(:civicrm_membership_status_id__0, :civicrm_membership_status_id__1, :civicrm_membership_status_id__2))) AND ((civicrm_contact.custom_2 = :civicrm_contact_custom_2)) LIMIT 21 OFFSET 0; Array ( [:civicrm_contact_custom_2] => Baker [:civicrm_membership_status_id__0] => 1 [:civicrm_membership_status_id__1] => 2 [:civicrm_membership_status_id__2] => 3 ) in Drupal\views\Plugin\views\query\Sql->execute() (line 1543 of core/modules/views/src/Plugin/views/query/Sql.php).
```
It appears that the custom field is being treated as if it were a column in the civicrm_contact table, which is not the case.

jackrabbithanna:
Skvare has started working on improved custom field Views integration. There should be a significant PR coming in the next month or so. We are considering porting the entire D7 CiviCRM Core Views handling of custom fields to D8. This could mean breaking changes to those already using CE for D8 Views with custom fields. Anyone interested in this subject should post here or reach out to me. Consider that in the D7, custom fields were treated as a separate table, implicitly joined to the appropriate entity type base table, without a Views Relationship.

jamie-tillman:
[...]

PR #210, opened 2020-04-17
Add support for custom fields views integration.

puresyntax71:


jackrabbithanna:
For issue https://github.com/eileenmcnaughton/civicrm_entity/issues/200

jitendrapurohit:
Came across this PR as we saw some problems with the custom fields included in views. I've applied this change and looks like it works on all our views for now. For existing drupal views - we need to re-add custom fields on them. Thanks for the PR @jackrabbithanna.

I'll test with some more fields and let you know if we notice any issues.

jackrabbithanna:
Thanks for testing @jitendrapurohit ! 
This PR is to bring feature equivalence for custom fields in Views similar to how it worked in D7 .. For sure, if custom fields were used in a Views already, they have to be re-added after this PR. I don't see a way around that, as this exposes the custom field group table, and joins that table to the main entity table.

This PR will allow:
filters to work
field aggregation to work
custom field groups allowing multiple values to work

We've tested with each custom field type / widget type combination.

jackrabbithanna:
@dsnopek I'm definitely interested in hearing your thoughts about this PR. This is one way to make all Views features work with custom fields, whether multi-valued groups or not .. so
click table header sorting works
field aggregation
filters
sorts
arguments

BUT .. as you mentioned in https://github.com/eileenmcnaughton/civicrm_entity/issues/200#issuecomment-614714931 

Are we losing anything doing it this way, as far as "tight Entity API integration?" as you see it? If so, what?
If you have concerns about this approach, I'd love to have a discussion to see how they could be addressed

jackrabbithanna:
We're working on a update for this PR, which is to make use of the EntityField plugin ... trying to make that work with click sorting, field aggregation, etc...But if we use EntityField as a base class for custom fields then I think that will get field formatters back ... hopefully this week for that update.

dsnopek:
My main qualm with this PR, is that I don't think we should be doing custom Views field plugins for different types of custom CiviCRM fields.

If a field needs custom logic in order to be displayed, we should handle that at the Entity Field API level, by making the field be a custom type with a custom field formatter to display it correctly. We need this same logic for displaying the field when rendering it on it's own page or whereever anyway! Then the Views field handler should just be about selecting and configuring the field formatter, reusing that same same logic, and meeting Drupal users expectations for how Entity fields work in Views.

Having custom Views filter, sort and relationship plugins all seems fine (and, in fact, necessary!)

I see now how this is breaking compatibility with the current civicrm_entity - it's moving the field definition from the entities base table (which was really a psuedo field anyway) to the custom field's table. When I was personally thinking about how to fix Views filters, I was imagining that we'd just put a custom filter handler on the same psuedo field on the base table which did the join to actual custom field table. It could still be done that way, but compatibility with existing Views from civicrm_entity isn't really a priority for me, so I'm not gonna argue that point. :-) I care much more about how the fields are displayed as described above.

jackrabbithanna:
Thank you for your feedback!

We're working an idea to address the main qualm, which I understand and agree with...I agree that if formatting is necessary, a field formatter is the way to go. It's not really the formatting that is the issue we want to solve anyway, it's the underlying query that views produces.

We're trying out a new field plugin that extends the EntityField plugin and overriding some of the methods to make aggregation / click sorting / multiple value handling work right

Custom fields is really a larger issue, as I mention in the issue that references this PR .. which we never made right in the D7 version. 
I think we may have a way forward though to get Views working, and I think in a way that future updates to base field definition and load/save for entity types with custom fields doesn't require a breaking change in the Views integration.

Hopefully this we week that Arnold pushes an update to this PR for your evaluation.

jackrabbithanna:
Arnold's posted a commit with a fairly big change, fields are now using EntityField or an extension of it. This what we're thinking for the more final solution.

bmango:
I have downloaded the latest code. Thank you very much! I have found a couple of issues: 

1. I am unable to add in a relationship between the user and Civi contact. When I go to the add relationship screen, I can see the relationship for `CiviCRM contact related to the user`, but when I tick the box and click 'Add and configure relationships' nothing happens. 

2. I did have a view built based on the updated code you had sent me with a relationship between the user and Civi Contact, however when I try and edit the view I get an error: 

`Uncaught PHP Exception Drupal\\Component\\Plugin\\Exception\\PluginException: "Plugin (civicrm_entity_civicrm_contact_user) instance class "Drupal\\civicrm_entity\\Plugin\\views\\relationship\\CiviCrmContactUser" does not exist." at /home/wfto/os.wfto.info/dir/html/core/lib/Drupal/Component/Plugin/Factory/DefaultFactory.php line 97, referer: http://os.wfto.info/admin/structure/views`

This last error doesn't matter to me as I can re-create the view, just thought you might like to see it.

jackrabbithanna:
I think those 2 issues are unrelated to the custom fields PR .. to get the user -> contact relationship there's https://github.com/eileenmcnaughton/civicrm_entity/pull/213

jackrabbithanna:
@puresyntax71 can you re-roll this?

jackrabbithanna:
We're using this PR on multiple builds .. and so far it's working out good. In general I think this PR is the way to go. Over time there may be little fixes / additions that need doing, but as far as "how to handle custom fields Views integration" .. I think this is it.

bmango:
Thanks very much for the latest version. I seem to be getting one error. When I inlude the contact name field in a user view, if I tick "_Link to the CiviCRM Contact_" in the field settings, then I get a "_The website encountered an unexpected error. Please try again later_." when looking at the view. In the Drupal log the error is: 

`Drupal\Core\Entity\Exception\UndefinedLinkTemplateException: No link template 'canonical' found for the 'civicrm_contact' entity type in Drupal\Core\Entity\EntityBase->toUrl() (line 227 of /home/.../html/core/lib/Drupal/Core/Entity/EntityBase.php)`

jackrabbithanna:
@bmango I think this issue is unrelated to custom fields. I am also not able to replicate this. For that, if you continue to have issues, please make a separate issue and we'll discuss it there.

bmango:
Ok, no problem.

jackrabbithanna:
I plan to merge this Monday unless there's a compelling argument, bad flaw, or compelling objection that gives reason not to. If there's little bugs we'll fix them as they come.

jackrabbithanna:
No objections, and nothing major bad coming out of our testing. Merging #210 
If bugs are found, please make a new issue in this repo issue queue and we will address them.

pradpnayak:
We ran into problems with 'Broken/missing handler' error on all views for custom field fields, filters, sorts etc after applying the patch. Had to update 40+ views manually. Have bunch of other sites using same module, hope so we dont have to update there as well.

Reason: previously the custom fields were stored as {{ custom_31 }} now as {{ image_link_31 }} (custom field id. 

Thanks
Pradeep

jackrabbithanna:
Yes that was necessary. That's the way it is. Won't have to do it again though.

PR #211, opened 2020-04-28
Fixes #198: Treat Address 'Country' like a pseudoconstant rather than…

### LINKED ISSUE #198: Country field doesn't render
On CiviCRM Address entities, the State/Province renders but the Country doesn't, even though these fields should be roughly identical in function.

As a sidenote, I tried troubleshooting this, but I don't know enough about how to force a call to `Drupal\civicrm_entity\Entity\FieldDefinitionProvider::getBaseFieldDefinition`.  I assume this is cached somehow, since it's not triggered when modifying a View; if you can give me a tip on where to look I can start fixing some of my own problems and sending PRs back.

dsnopek:
… an entity reference.

I'm not sure this is the right thing to do - see #198 for the details.

### DIFF
--- a/src/SupportedEntities.php
+++ b/src/SupportedEntities.php
@@ -171,18 +171,6 @@ final class SupportedEntities {
         'delete' => ['access CiviContribute', 'administer CiviCRM'],
       ],
     ];
-    $civicrm_entity_info['civicrm_country'] = [
-      'civicrm entity label' => t('Country'),
-      'civicrm entity name' => 'country',
-      'label property' => 'name',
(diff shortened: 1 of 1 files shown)


KarinG:
I agree - I would not want to manage/create Countries or edit them (move them to another continent/region). This PR works and I think this solution [that we don't make Country an Entity] makes more sense -> here's the countries 👍 

image

jackrabbithanna:
I think we can close this in favor of 
https://github.com/eileenmcnaughton/civicrm_entity/pull/212

PR #214, opened 2020-05-14
Civi Address leaflet map submodule

jackrabbithanna:
Submodule providing a Leaflet map Views format style plugin. 

Instead of needing a geofield, the user is able to choose fields from the View which have the lat/lon pair. Intended for use with CiviCRM address geo_code_1 and geo_code_2 properties. 

Just add those two fields to a view, set the new View format, and match up the fields to the lat/lon settings. 

@KarinG this is the same module I shared with you.

I could make this a separate module project but I figured this would be a good thing to bundle with CE

We've got a Views proximity filter coming soon and goes along well with a nice Leaflet map!


KarinG:
I tested this and yeah - this works as advertised 👍 

I do like the idea of bundling it with D8CE.

jackrabbithanna:
Merging #214

PR #222, opened 2020-07-20
Allow NULL.

puresyntax71:


jackrabbithanna:
Coming from testing and actually usage on Skvare projects, merging #222

PR #225, opened 2020-08-04
Disable hooks feature

puresyntax71:


jackrabbithanna:
This adds an "Advanced Settings" fieldset to the CiviCRM Entity settings page. Adds a checkbox that will disable the hook_civicrm_pre() and hook_civicrm_post() implementations. 

We had run into some situations, both with large data import processes, which caused performance issues, and large PHP memory usage. 

One was a D7 -> D8 migration of users
Another was importing 100s of thousands of Contacts/memberships via a CiviCRM API based importer. 

Good to have, most people will never need to enable it though.

Merging #225

PR #226, opened 2020-08-04
Views schema for proximity filter

puresyntax71:


jackrabbithanna:
Couple fixes we have for the new proximity Views filter. Tested in a production environment. Merging #226

PR #227, opened 2020-08-04
Allow CiviCRM extensions to alter supported entities

puresyntax71:


jackrabbithanna:
Adds ability for modules/extensions to alter the supported CiviCRM entity metadata. Merging #227

Issue #220, opened 2020-07-08
Drupal 8, Multilingual Instance using views


seamuslee001:
When running a Drupal 8 site in multilingual mode e.g. en_US and fr_CA as we are doing for a Canadian Client, if you have a view that queries events, and you ask to pull out the title or description fields, these currently break because the CiviCRM entity view will attempt to query the table civicrm_event and try and select civicrm_event.title for example however the title column has been dropped and replaced with 2 new columns in this case but it is 1 column per locale the columns are then title_<locale> e.g. title_en_US and title_fr_CA

In Drupal 7 the core integration for views fixed this by re-writing the query

https://github.com/civicrm/civicrm-drupal/blob/a17cfdf12b93d52b98197b048383b8b65ce72a23/civicrm.module#L787

For our client we have so far done the following

```php
function jma_customizations_views_query_alter(ViewExecutable $view, QueryPluginBase $query) {
  if ($view->id() == 'events') {
    // Traverse through the 'where' part of the query.
    foreach ($query->where as &$condition_group) {
      foreach ($condition_group['conditions'] as &$condition) {
        // If this is the part of the query filtering on title, chang the
        // condition to filter on node ID.
        if ($condition['field'] == 'civicrm_event.title') {
          $condition['field'] = 'civicrm_event.title_en_US';
        }
        if ($condition['field'] == 'civicrm_event.description') {
          $condition['field'] = 'civicrm_event.description_en_US';
        }
     }
   }
  }
}
```
However I would say that i think something along the lines of what the core module had done would be most sensible right? @jackrabbithanna

jackrabbithanna:
Yes, I like what you're thinking there. Updates along these lines seem appropriate to me.

jackrabbithanna:
Fixes included in -dev release, and beta2 tomorrow

PR #229, opened 2020-08-09
#220 Alter views query to match the related translated columns when running on multilingual database

seamuslee001:
ping @jackrabbithanna 

I noticed that I also got sometimes the $condition['field'] was an object of type Condition, which i suspect we may need to work out how to transverse completely but at the same time i think this will at least be a decent improvement


jackrabbithanna:
For sure doesn't break Views for those people not using multi-lingual. Merging #229

PR #230, opened 2020-08-10
Include Option Value as a supported entity

seamuslee001:
ping @jackrabbithanna 

The use case was I was building a search form in an app and using graphql to pull out the option values for a specific option group, without this couldn't query the option values


jackrabbithanna:
Thanks Seamus, we're testing this out for any unintended consequences and will merge if it survives some battle testing.

jackrabbithanna:
Doesn't affect existing functionality in a way that I can see, and opens up new possibilities. Merging #230

PR #231, opened 2020-08-10
Update hook signature to be non deprecated format

seamuslee001:
as per https://github.com/civicrm/civicrm-core/pull/17995 This updates the hook signature to use the non deprecated format for calling a hook ping @jackrabbithanna


jackrabbithanna:
Thanks Seamus .. we're testing this now, sure to merge soon.

jackrabbithanna:
This worked for us. merging #231

PR #238, opened 2020-09-27
Update Hook signatures when invoking hooks

ghost:
The signature used when calling CRM_Utils_Hook->invoke is deprecated, and from CiviCRM 5.29 this gives rise to a log warning message every time it occurs.  This change updates the signature to the preferred version and cleans up the log.  

I have guessed the parameter names to be used in the three cases since I can't find documentation or examples to say what they should be.  The hooks are called successfully, but there may be other subtle side effects I am not aware of.

It is applied to CiviCRM Versions >= 5.28 because I do not know the backward compatibility situation.  Would a better solution be to discard the old deprecated versions and update the compatibility statement in the info.xml file?

### DIFF
--- a/civicrm_entity.module
+++ b/civicrm_entity.module
@@ -488,7 +488,14 @@ function civicrm_entity_get_supported_entity_info($entity_type = NULL) {
   if (civicrm_initialize()) {
     $codeVersion = explode('.', CRM_Utils_System::version());
     // if db.ver < code.ver, time to upgrade
-    if (version_compare($codeVersion[0] . '.' . $codeVersion[1], 4.5) >= 0) {
+    if (version_compare($codeVersion[0] . '.' . $codeVersion[1], 5.28) >= 0) {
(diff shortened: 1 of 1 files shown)


seamuslee001:
This looks fine to me @eileenmcnaughton @jackrabbithanna

jackrabbithanna:
@seamuslee001  so this PR would be preferable to this other one? https://github.com/eileenmcnaughton/civicrm_entity/pull/244/files

jackrabbithanna:
Merging on Seamus's recommendation, Merging #238

edvanleeuwen:
Tested and verified.

Upperholme:
I don't suppose this is going to make an appearance in a 7.x release from d.o ?

somnolentsurfer:
I've just updated to CiviCRM 5.34 and I'm getting this error displayed a load in the UI, not just in the logs. Any chance there'll be release the includes this fix for Drupal 7 any time soon?

olivierh65:
Installed and there is more warnings.

PR #239, opened 2020-10-02
initial commit

puresyntax71:


jackrabbithanna:
Adds a relationship plugin, for relating Contacts to Activities, or Activities to contacts.
Removes the 3 auto generated ones that were there, and didn't work. 
Relationship plugin offers options for Source, Target, or Assignee contacts .. so the record_type_id of the civicrm_activity_contact table. 
"Any" option is also available. 

Merging #239

PR #241, opened 2020-10-11
8.x 3.x views arg default contact subtype

rob1263:
Adds a views default argument plugin to allow for a contacts view contextual filter that returns only contacts who are in the same civi subtype as the logged in user.

Based on the default arg plugin for logged in user's uid.

### DIFF
--- a/src/Plugin/views/argument_default/ContactSubtype.php
+++ b/src/Plugin/views/argument_default/ContactSubtype.php
@@ -23,7 +23,7 @@ class ContactSubtype extends ArgumentDefaultPluginBase implements CacheableDepen
    */
   public function getArgument() {
 
-    $current_user_contact_subtype = '';
+    $current_user_contact_subtype = '<none>';
 
     $civicrm_api = \Drupal::getContainer()->get('civicrm_entity.api');                                                 
 
--- a/src/Plugin/views/argument_default/ContactSubtype.php
+++ b/src/Plugin/views/argument_default/ContactSubtype.php
@@ -5,6 +5,9 @@ namespace Drupal\civicrm_entity\Plugin\views\argument_default;
 use Drupal\Core\Cache\Cache;
 use Drupal\Core\Cache\CacheableDependencyInterface;
 use Drupal\views\Plugin\views\argument_default\ArgumentDefaultPluginBase;
+use Symfony\Component\DependencyInjection\ContainerInterface;
+use Drupal\Core\Session\AccountProxy;
+use Drupal\civicrm_entity\CiviCrmApiInterface;
 
 /**
  * Default argument plugin to extract the current user's civicrm contact subtype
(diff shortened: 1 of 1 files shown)


KarinG:
Hi @jackrabbithanna - this adds an additional config/option for Contextual Filter, so much like we can make a My Activities we can now easily create a My Contacts view -> showing each chapter/club admin only those Contact that have the same Contact Subtype as the Logged in user:

image

image

KarinG:
@jackrabbithanna - do you think this makes sense? Happy to rework it some more if you can think of ways to make this more useful.

jackrabbithanna:
I think it would be best if the civicrm_entity.api service was made available via dependency injection. See this plugin as an example that uses the same base class: https://api.drupal.org/api/drupal/core%21modules%21taxonomy%21src%21Plugin%21views%21argument_default%21Tid.php/8.9.x
The create() and __construct() method. 
That's best practice for using services in a Drupal plugin
If that was done, instead of 
$civicrm_api = \Drupal::getContainer()->get('civicrm_entity.api'); 
then I'd merge this.

Other thoughts, but not deal breakers. 
This would not support contacts with multiple sub types, or it would always use the first one in the array retrieved from the API
It's very rare in my experience to have multiple sub-types for a contact ..so edge-casey .. but worth noting and if could be made to work that would be superb.

It's also rare that a contact's sub types would change, but if it did cache may should be invalidated. This I'm not going to worry about that for this PR in particular, but civicrm entity cache tags has been on my mind to explore some day.

KarinG:
Thanks @jackrabbithanna - we'll get back to you soon.

rob1263:
Thanks @jackrabbithanna, I changed it to use dependency injection for the services.

Now I'm working on better handling of either zero or multiple contact subtypes for the logged in user. That's close, but I wanted to give the user the choice to do 'AND' or 'OR' for multiple subtypes and I can't get that to work. Whether I use a comma or a plus to delimit the subtypes in the views argument, it always does an OR.

This isn't just a problem with this new plugin. For instance, I tested with a contextual filter for contact subtype, choosing the existing 'Fixed' argument for the default value, and entering a comma delimited list of subtypes (being sure to set 'multiple values' for the filter), and it always does OR instead of AND.

It is parsing the delimited string, because OR does work. I'm still looking into it, but thought I'd post about it here in case you had another suggestion. :)

Thanks.

jackrabbithanna:
I've never done it personally. But I can point to an example in Drupal core, that TID argument default plugin
The form options:
https://git.drupalcode.org/project/drupal/-/blob/8.9.x/core/modules/taxonomy/src/Plugin/views/argument_default/Tid.php#L153

Then in getArguments() method
https://git.drupalcode.org/project/drupal/-/blob/8.9.x/core/modules/taxonomy/src/Plugin/views/argument_default/Tid.php#L213

They implode the array by either '+', or ',' .. depending on the user choice. Looks like '+' = OR and =',' = AND

So it would seem to be that you'd implode 
$results[0]['contact_sub_type'] by that option. 

That's my best guess

rob1263:
Thanks. It seems like the comma delimited AND not doing the right thing is maybe a problem for the core NumericArgument.php too...
https://www.drupal.org/project/drupal/issues/3061546

So I'm just going to ignore that for now, since it's not a case we need to support yet.

I added options for handling zero user subtypes (filter to return none or 'all') or multiple subtypes (use first subtype or match any (OR)).

Re the cache, I've been testing and changing user subtypes while working on this, and haven't had to clear caches to see changes in the results. So maybe the argument's not getting cached, or the cache context 'user' is taking care of it somehow? I have no idea how that works. :)

jackrabbithanna:
Ok .. this is great. Thanks for contributing! Got something good and if people want the more and the better, we'll do it then. 

Merging #241

KarinG:
Thank you!

PR #247, opened 2020-10-23
8.x 3.x - D9 Compatibility

stesi561:
Summary of changes:

Bumps drupal core version to 8.8^ and changes to core_version_requirement syntax.

Removes depreciated functions and classes replacing these appropriately

### DIFF
--- a/civicrm_entity.info.yml
+++ b/civicrm_entity.info.yml
@@ -3,6 +3,7 @@ type: module
 description: 'Expose CiviCRM entities as Drupal entities.'
 package: CiviCRM
 core: 8.x
+core_version_requirement: ^8.7.7 || ^9
 dependencies:
   - civicrm
   - drupal:datetime
--- a/civicrm_entity.info.yml
+++ b/civicrm_entity.info.yml
@@ -2,7 +2,7 @@ name: CiviCRM Entity
 type: module
 description: 'Expose CiviCRM entities as Drupal entities.'
 package: CiviCRM
-core_version_requirement: ^8.7.7 || ^9
+core_version_requirement: ^8.8 || ^9
 dependencies:
   - civicrm
   - drupal:datetime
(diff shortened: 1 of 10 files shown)


jackrabbithanna:
Very nice .. Thanks for contributing. Will be testing this PR with Drupal 8 to make sure it gels, mostly looks very simple and obvious. I think if we merged this then people would need D8.8 or above. Decision time, do we maintain backward compatibility with older D8 versions, and start a 4.x branch...or do we merge this and people upgrade. I wonder if many people use pre 8.7.7 anymore.
I'm fine with only supporting 8.8 and above moving forward...

KarinG:
Awesome Luke! Yes Mark -> only for D8.8 and up! 👍

stesi561:
One option might be to move to a semantic versioning -> https://www.drupal.org/node/1015226#semver-transition

stesi561:
I've done basic testing - install + can I create a view of civicrm contacts on Drupal 8 and 9.

If you are testing - don't try and uninstall - it will wipe your civicrm database.

In case it helps - I had some fun with composer to get it installing correctly - which lead to realisation that the order matters on the repository section!

To test this using composer you can add the following to your composer.json repository section ABOVE the 'type': 'composer' item.

i.e.
```
 "repositories": [
 	{
 		"type": "vcs",
 		"url": "https://github.com/stesi561/civicrm_entity"
 	},
        {
            "type": "composer",
            "url": "https://packages.drupal.org/8"
        }
    ],

```

Then you can 
```
composer require drupal/civicrm_entity:dev-8.x-3.x
```

jackrabbithanna:
We had been working in parallel on some D9 compatibility fixes. Much of what is in this PR was in ours. 
What I'm going to do is merge this PR .. then we will follow up in the next day or two with a few small tweaks and additions.
Merging #247

stesi561:
Thanks @jackrabbithanna Happy to have some other eyes on this!

I'll have a look at #248 in the next day or so.

jackrabbithanna:
Great! Thank you!
We have been and will be doing the same. 
If we all do a little testing on the github repo version .. I can feel comfortable about releasing beta2 on D.O
Not expecting any issues ATM, but nobody ever does.

PR #250, opened 2020-11-02
Add proximity fixes.

puresyntax71:


jackrabbithanna:
Unfortunately we had trouble with Ajax loading of state provinces when use selects country, either in the filter settings, or in an exposed filter. We made the decision to hardcode country to USA, I understand others may have desire for other countries. 
This issue https://www.drupal.org/project/drupal/issues/2842525 appears to be the problem. 
The fixes are necessary for it to work at all, so I'm merging, knowing that it's not perfect yet.

PR #253, opened 2020-11-06
Add mutliple checkbox.

puresyntax71:


jackrabbithanna:
This is to make the field metadata for custom fields that have 'Checkboxes' Input type have UNLIMITED cardinality (multiple values) on the Drupal side of the entity type definitions. 

This helps with these fields on the Drupal version of CiviCRM entity type add/edit forms

Merging #253

PR #254, opened 2020-11-09
Add a way to ignore mismatch definitions.

puresyntax71:


seamuslee001:
I applied this to a client site and it correctly removed the status notification about the mismatched field schema @KarinG @jackrabbithanna

jackrabbithanna:
Merging #254

PR #255, opened 2020-11-09
Prevent data loss on uninstall.

puresyntax71:
This decorations `ModuleInstaller` although this removes all validators for `civicrm_entity`. I think it would be nice to be able to just have it ignore the `ContentUninstallValidator` instead since it's the only validator that is responsible for validating content existence for `civicrm_entity`.


jackrabbithanna:
@puresyntax71 I believe this can be taken out of WIP status. 

We have this running now on our demo sites, and some development instances soon to go to production

jackrabbithanna:
Merging #255 .. this will allow people to be able to uninstall without wrecking their CiviCRM database .. so want it in beta2 release

KarinG:
👍 Awesome!

PR #258, opened 2020-12-14
Fix variable overwrite into civicrm_entity_views_query_alter

GValFr35:
Fix PHP warnings when calling array_key_exists () on string.
This is due to an array ($columns) which is replaced by a string during the previous loop.


jackrabbithanna:
Thanks for the contribution. Probably to merge,  in our testing queue now.

PR #260, opened 2020-12-16
Fixes a module loading fix in D7 (PHP 7.3 compatibility)

reecebenson:
This change fixes an issue when loading modules with Drupal 7, in our case, it was making the site's token module fall over.

> Warning: "continue" targeting switch is equivalent to "break". Did you mean to use "continue 2"? in require_once() (line 341 of /home/[redacted]/public_html/includes/module.inc).

Lead back to this module, and this file. Updating this line to be `continue 2` resolved the issue for us.


jackrabbithanna:
Thanks for contribution, merging #260

KarinG:
Noting that there are more instances -> this one e.g. 

```
root@fcd898a3d465:/var/www/html/sites/default/modules/civicrm_entity# sudo -u www-data drush cc all
"continue" targeting switch is equivalent to "break". Did you mean to use "continue 2"? date_views_fields.inc:86[warning]
'all' cache was cleared.                                                                                        [success]
```
 
PR coming today

KarinG:
Ah no PR needed -> fixed here: https://www.drupal.org/project/date/issues/3009740

PR #262, opened 2020-12-22
Fix entity permission for related contact

jitendrapurohit:
Views having related contact does not respect view my contact + relationship permission. 

To replicate

- Build a view on CiviCRM Contact.
- Add a relationship to civicrm_relationship table and display the related contact B on the view result.
- Add a contextual filter for current user id (contact A).
- Make sure contact A only has `view my contact` permission + civi relationship permission to view and update contact B.
- Login as contact A and navigate to the view result. The contact B is not displayed to the user even if it has the required permission. 

Here's the view config from our setup - https://gist.githubusercontent.com/jitendrapurohit/9e35965342e76b5f39f3d8074aceb4c5/raw/ddab9936487f5bb880fac770740e711738e95276/relationship-view-config

### DIFF
--- a/src/CivicrmEntityAccessHandler.php
+++ b/src/CivicrmEntityAccessHandler.php
@@ -72,6 +72,13 @@ class CivicrmEntityAccessHandler extends EntityAccessControlHandler {
     if (!empty($this->civicrmEntityInfo[$this->entityTypeId]['permissions'][$operation])) {
       $permissions = $this->civicrmEntityInfo[$this->entityTypeId]['permissions'][$operation];
     }
+    if ($this->entityTypeId == 'civicrm_contact' && in_array($operation, ['view', 'edit'])) {
+      \Drupal::service('civicrm')->initialize();
+      $op = $operation == 'view' ? \CRM_Core_Permission::VIEW : \CRM_Core_Permission::EDIT;
+      if (\CRM_Contact_BAO_Contact_Permission::allow($entity->id(), $op)) {
+        return AccessResult::allowed();
+      }
+    }
     return AccessResult::allowedIfHasPermissions($account, $permissions, 'OR');
   }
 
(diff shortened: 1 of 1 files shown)


jackrabbithanna:
Thanks for contributing, in our testing queue now to get in before beta2 release

jitendrapurohit:
Have pushed another related update that fixes the loading of related contacts membership. Eg 

- Contact A is the primary member with a related/inherited membership to contact B.
- Consider a view that loads the logged-in user membership data. (member id, status, start date, end date, primary member id, etc)
- As contact B - the current code does not load the value in the primary member id column.

jackrabbithanna:
I have trouble replicating the original issue. 
I have test user, with only 'view my contact' permission
Test user has "Employee of" relationship, with "View and update" permission, to an org contact

I created a View, base table civicrm_contact with relationships:
CiviCRM Relationship using contact_id_a .. to join to civicrm_relationship table, civicrm_contact.id = civicrm_relationship.contact_id_a
CiviCRM Contact (id of the second contact)

I then add 2 fields, display name, and display name again, using the "id of second contact" relationship

When I look at the View page, the test user is able to see the the display name of the related contact. 
Using CiviCRM 5.27

I also tried adding the related user relationship, and contextual filter of user id, using default value of logged in user.

jackrabbithanna:
I've tried to import your View on two sites .. I get error trying to import. Not sure why. 

Here's a gist of the View I'm building to test this:
https://gist.github.com/jackrabbithanna/503370c64a07120c3544445cff5b3697

jackrabbithanna:
Re-considering this PR .. will be reviewing and testing for possible inclusion to the module

gibsonoliver:
We had a situation using Views and Drupal 11, contact data was not showing on public page displays. This patch helped.

PR #263, opened 2021-01-04
Entity always validated as an event type

mglaman:
This ensures the entity is validated based on its actual type.


jackrabbithanna:
Merging #263

PR #264, opened 2021-01-08
Add a computed end date for activities

mglaman:
This adds a computed field to provide an end date for Activities. This takes the start date and time and adds the duration to make a computed value. This has been tested with the [Fullcalendar View](https://www.drupal.org/project/fullcalendar_view). 

This also supports updating the duration with Fullcalendar when the event length is dragged.

### DIFF
--- a/src/Plugin/Field/ActivityEndDateFieldItemList.php
+++ b/src/Plugin/Field/ActivityEndDateFieldItemList.php
@@ -3,13 +3,22 @@
 namespace Drupal\civicrm_entity\Plugin\Field;
 
 use Drupal\civicrm_entity\Entity\CivicrmEntity;
-use Drupal\Core\Datetime\DrupalDateTime;
 use Drupal\Core\Field\FieldItemList;
 use Drupal\Core\TypedData\ComputedItemListTrait;
 use Drupal\datetime\Plugin\Field\FieldType\DateTimeItemInterface;
 
 /**
  * A computed field item list for Activities to provide an end date and time.
+ *
+ * In this class, you will notice timezone conversions from the default timezone
+ * to UTC. CiviCRM stores dates in the expected timezone, and Drupal always
+ * stores them in UTC. So we have to do a bit of conversion.
+ *
+ * This is also automatically done when a CivicrmEntity instance is saved and
+ * when an entity is loaded.
+ *
+ * @see \Drupal\civicrm_entity\Entity\CivicrmEntity::civicrmApiNormalize
+ * @see \Drupal\civicrm_entity\CiviEntityStorage::initFieldValues
  */
 class ActivityEndDateFieldItemList extends FieldItemList {
   use ComputedItemListTrait;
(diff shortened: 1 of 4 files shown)


src/Plugin/Field/ActivityEndDateFieldItemList.php
```diff
@@ -0,0 +1,56 @@
...
+    $entity = $this->getEntity();
+    assert($entity instanceof CivicrmEntity);
+    $activity_date_time = strtotime($entity->get('activity_date_time')->value);
+    $new_end_date = strtotime($this->get($delta)->value);
+    $diff = $new_end_date - $activity_date_time;
+    // Update the duration of the activity.
+    $minutes = $diff / 60;
+    $entity->get('duration')->setValue($minutes);
+    parent::onChange($delta);
+  }
```
mglaman: This allows updating the duration by manipulating the end date via Fullcalendar

https://user-images.githubusercontent.com/3698644/103976599-cb8f8400-513c-11eb-91b8-0cf7b4a81269.mov

KarinG:
🆒 Testing this now!

KarinG:
Before PR
image

After PR :-) Now we have a timezone issue I think -> all end dates look like they are UTC now:
image

KarinG:
Confirming changing lunch to 12-2pm -> become 12-9pm (so + 7th - which is UTC - MST). Will give it a closer look in a bit!

mglaman:
Okay, I probably need to pass the timezone into those date objects

KarinG:
Yes -> I think:

`$timezone = drupal_get_user_timezone();`

Trying that now!

KarinG:
We just hit another bug in webform civicrm module (contacts/relationships) - that I'm going to need to look at first. 

@mglaman - handing this back to you! If you can work in: 
`$timezone = drupal_get_user_timezone();`

That would be great!

mglaman:
I wonder if this is also the culprit

```
    $activity_date_time = strtotime($entity->get('activity_date_time')->value);
    $new_end_date = strtotime($this->get($delta)->value);
```

Maybe we have to construct date objects since the duration is what has gone out of whack.

KarinG:
Yes - start date/time is still 👍

mglaman:
😅 so I found my problem. "Users may set their own time zone" was turned on, and the admin user was set to UTC whereas the site and server were America/Chicago.

\Drupal\system\TimeZoneResolver::setDefaultTimeZone runs at the beginning of the request and when the logged in user is set to modify the default timezone via `date_default_timezone_set`.

KarinG:
Ah ok! And on our end we don't "allow Users to set their own time zone" (it's a provincial non-profit).

mglaman:
Okay, this was _a bit_ tricky. I made sure to comment that in the code.

In `\Drupal\civicrm_entity\CiviEntityStorage::initFieldValues` we have the following code for timezone handling:

```
          else {
            $datetime_format = $definition->getSetting('datetime_type') === DateTimeItem::DATETIME_TYPE_DATE ? DateTimeItemInterface::DATE_STORAGE_FORMAT : DateTimeItemInterface::DATETIME_STORAGE_FORMAT;
            // CiviCRM gives us the datetime in the users timezone (or no
            // timezone at all) but Drupal expects it in UTC. So, we need to
            // convert from the users timezone into UTC.
            $datetime_value = (new \DateTime($item[$main_property_name], new \DateTimeZone(date_default_timezone_get())))->setTimezone(new \DateTimeZone('UTC'))->format($datetime_format);
            $item_values[$delta][$main_property_name] = $datetime_value;
          }
```

I didn't think this was being called, but it turns out it was. I assumed it was only called on the first load, or maybe it is on save, and that caused the duration to berserk out because it is calculated. I was basically performing the same logic. So it went from Chicago -> UTC -> UTC -> Chicago.

This _should_ tidy it up.

src/Plugin/Field/ActivityEndDateFieldItemList.php
```diff
@@ -0,0 +1,69 @@
...
+    if (!$activity_date_time) {
+      return;
+    }
+    // The time is already in UTC due to ::initFieldValues in storage.
+    // @see \Drupal\civicrm_entity\CiviEntityStorage::initFieldValues
+    $date = new \DateTime($activity_date_time, new \DateTimeZone('UTC'));
+    // We have to change this _back_ to the default timezone, as initFieldValues
+    // will be called again and it assumes the value is from CiviCRM, in the
+    // default timezone.
+    $date->setTimezone(new \DateTimeZone(date_default_timezone_get()));
```
mglaman, line 43: This is the workaround. When we computed the end date, we take the time as it is in UTC, convert it to the default timezone. We pass the time in this timezone to the field because `initFieldValues` will push it back to UTC for us.

KarinG:
This is totally working now! 🎉👏
(Couldn't wait until tomorrow - had to try!)

jackrabbithanna:
@KarinG whenever you're happy with this we can merge.

KarinG:
Mark: I'm SUPER happy with this. This is massive for scheduling. Live tested!

PR #265, opened 2021-01-09
Add support for bundles

mglaman:
We can have bundles on entities without there needing to be a UI for fields or anything else. 

Doing so would allow for some typical Drupal module integrations that expect bundles. For instance, with Fullcalendar View I can specify different colors per bundle.

This ties into #264 

Screen Shot 2021-01-08 at 10 30 34 PM

### DIFF
--- a/civicrm_entity.module
+++ b/civicrm_entity.module
@@ -125,8 +125,10 @@ function civicrm_entity_entity_type_build(array &$entity_types) {
     if (!empty($civicrm_entity_info['bundle property'])) {
       $entity_type_info['entity_keys']['bundle'] = 'bundle';
       $entity_type_info['civicrm_bundle_property'] = $civicrm_entity_info['bundle property'];
-      $entity_type_info['links']['add-page'] = $entity_type_info['links']['add-form'];
-      $entity_type_info['links']['add-form'] = sprintf('%s/{%s}', $entity_type_info['links']['add-form'], $entity_type_info['entity_keys']['bundle']);
+      if (isset($entity_type_info['links'])) {
+        $entity_type_info['links']['add-page'] = $entity_type_info['links']['add-form'];
+        $entity_type_info['links']['add-form'] = sprintf('%s/{%s}', $entity_type_info['links']['add-form'], $entity_type_info['entity_keys']['bundle']);
+      }
     }
 
     $entity_types[$entity_type_id] = new ContentEntityType($entity_type_info);
--- a/civicrm_entity.module
+++ b/civicrm_entity.module
@@ -83,17 +83,6 @@ function civicrm_entity_entity_type_build(array &$entity_types) {
       ],
     ];
 
-    // If this entity has bundle support, we define the bundle field as "bundle"
-    // and will use the "bundle property" as the field to fetch field options
-    // from CiviCRM with.
-    //
-    // @see civicrm_entity_entity_bundle_info()
-    // @see \Drupal\civicrm_entity\Entity\CivicrmEntity::baseFieldDefinitions()
-    if (!empty($civicrm_entity_info['bundle property'])) {
-      $entity_type_info['entity_keys']['bundle'] = 'bundle';
-      $entity_type_info['civicrm_bundle_property'] = $civicrm_entity_info['bundle property'];
-    }
-
     if (in_array($entity_type_id, $enabled_entity_types)) {
       $entity_type_info = array_merge_recursive($entity_type_info, [
         'handlers' => [
@@ -120,6 +109,20 @@ function civicrm_entity_entity_type_build(array &$entity_types) {
         'field_ui_base_route' => "entity.$entity_type_id.collection",
       ]);
     }
+
+    // If this entity has bundle support, we define the bundle field as "bundle"
+    // and will use the "bundle property" as the field to fetch field options
+    // from CiviCRM with.
+    //
+    // @see civicrm_entity_entity_bundle_info()
+    // @see \Drupal\civicrm_entity\Entity\CivicrmEntity::baseFieldDefinitions()
+    if (!empty($civicrm_entity_info['bundle property'])) {
+      $entity_type_info['entity_keys']['bundle'] = 'bundle';
+      $entity_type_info['civicrm_bundle_property'] = $civicrm_entity_info['bundle property'];
+      $entity_type_info['links']['add-page'] = $entity_type_info['links']['add-form'];
+      $entity_type_info['links']['add-form'] = sprintf('%s/{%s}', $entity_type_info['links']['add-form'], $entity_type_info['entity_keys']['bundle']);
+    }
+
     $entity_types[$entity_type_id] = new ContentEntityType($entity_type_info);
   }
 }
--- a/civicrm_entity.module
+++ b/civicrm_entity.module
@@ -83,6 +83,18 @@ function civicrm_entity_entity_type_build(array &$entity_types) {
         'storage_schema' => CivicrmEntityStorageSchema::class,
       ],
     ];
+
+    // If this entity has bundle support, we define the bundle field as "bundle"
+    // and will use the "bundle property" as the field to fetch field options
+    // from CiviCRM with.
+    //
+    // @see civicrm_entity_entity_bundle_info()
+    // @see \Drupal\civicrm_entity\Entity\CivicrmEntity::baseFieldDefinitions()
+    if (!empty($civicrm_entity_info['bundle property'])) {
+      $entity_type_info['entity_keys']['bundle'] = 'bundle';
+      $entity_type_info['civicrm_bundle_property'] = $civicrm_entity_info['bundle property'];
+    }
+
     if (in_array($entity_type_id, $enabled_entity_types)) {
       $entity_type_info = array_merge_recursive($entity_type_info, [
         'handlers' => [
@@ -117,20 +129,25 @@ function civicrm_entity_entity_type_build(array &$entity_types) {
  * Implements hook_entity_bundle_info().
  */
 function civicrm_entity_entity_bundle_info() {
-  $bundles = [
-    'civicrm_activity' => [],
-  ];
-
   $transliteration = \Drupal::transliteration();
   /** @var \Drupal\civicrm_entity\CiviCrmApiInterface $civicrm_api */
   $civicrm_api = \Drupal::service('civicrm_entity.api');
-  $options = $civicrm_api->getOptions('activity', 'activity_type_id');
-  foreach ($options as $option) {
-    $machine_name = $transliteration->transliterate($option, LanguageInterface::LANGCODE_DEFAULT, '_');
-    $machine_name = mb_strtolower($machine_name);
-    $machine_name = preg_replace('/[^a-z0-9_]+/', '_', $machine_name);
-    $machine_name = preg_replace('/_+/', '_', $machine_name);
-    $bundles['civicrm_activity'][$machine_name]['label'] = $option;
+
+  $bundles = [];
+  $entity_types_with_bundles = array_filter(SupportedEntities::getInfo(), static function(array $civicrm_entity_info) {
+    return !empty($civicrm_entity_info['bundle property']);
+  });
+  foreach ($entity_types_with_bundles as $entity_type_id => $civicrm_entity_info) {
+    $bundles[$entity_type_id] = [];
+    $options = $civicrm_api->getOptions($civicrm_entity_info['civicrm entity name'], $civicrm_entity_info['bundle property']);
+    foreach ($options as $option) {
+      $machine_name = $transliteration->transliterate($option, LanguageInterface::LANGCODE_DEFAULT, '_');
+      $machine_name = mb_strtolower($machine_name);
+      $machine_name = preg_replace('/[^a-z0-9_]+/', '_', $machine_name);
+      $machine_name = preg_replace('/_+/', '_', $machine_name);
+
+      $bundles[$entity_type_id][$machine_name]['label'] = $option;
+    }
   }
   return $bundles;
 }
--- a/civicrm_entity.module
+++ b/civicrm_entity.module
@@ -17,10 +17,15 @@ use Drupal\civicrm_entity\Routing\CiviCrmEntityRouteProvider;
 use Drupal\civicrm_entity\SupportedEntities;
 use Drupal\Core\Entity\ContentEntityDeleteForm;
 use Drupal\Core\Entity\ContentEntityType;
+use Drupal\Core\Entity\Display\EntityViewDisplayInterface;
+use Drupal\Core\Entity\EntityDisplayRepositoryInterface;
+use Drupal\Core\Entity\EntityTypeInterface;
 use Drupal\Core\Entity\FieldableEntityInterface;
 use Drupal\Core\Field\FieldStorageDefinitionInterface;
 use Drupal\Core\Render\Element;
 use Drupal\Core\StringTranslation\TranslatableMarkup;
+use Drupal\field\Entity\FieldConfig;
+use Drupal\field\FieldConfigInterface;
 use Drupal\views\ViewExecutable;
 use Drupal\views\Plugin\views\query\QueryPluginBase;
 
@@ -140,7 +145,13 @@ function civicrm_entity_entity_bundle_info() {
     return !empty($civicrm_entity_info['bundle property']);
   });
   foreach ($entity_types_with_bundles as $entity_type_id => $civicrm_entity_info) {
-    $bundles[$entity_type_id] = [];
+    // We keep a bundle that is the same as the entity type ID. This allows us
+    // to create fields as if this entity has no bundles.
+    $bundles[$entity_type_id] = [
+      $entity_type_id => [
+        'label' => $civicrm_entity_info['civicrm entity label'],
+      ],
+    ];
     $options = $civicrm_api->getOptions($civicrm_entity_info['civicrm entity name'], $civicrm_entity_info['bundle property']);
     foreach ($options as $option) {
       $machine_name = SupportedEntities::optionToMachineName($option, $transliteration);
@@ -148,6 +159,61 @@ function civicrm_entity_entity_bundle_info() {
     }
   }
   return $bundles;
+}
+
+/**
+ * Implements hook_entity_bundle_field_info().
+ *
+ * This ensures CiviCRM Entity entity types have their field config instances
+ * across all bundles. It's a copy of the Field module's logic, but clones
+ * field config definitions.
+ *
+ * @see field_entity_bundle_field_info().
+ */
+function civicrm_entity_entity_bundle_field_info(EntityTypeInterface $entity_type, $bundle, array $base_field_definitions) {
+  $result = [];
+  if ($entity_type->get('civicrm_entity_ui_exposed') && $entity_type->hasKey('bundle')) {
+    // Query by filtering on the ID as this is more efficient than filtering
+    // on the entity_type property directly.
+    $ids = \Drupal::entityQuery('field_config')
+      ->condition('id', $entity_type->id() . '.', 'STARTS_WITH')
+      ->execute();
+    // Fetch all fields and key them by field name.
+    $field_configs = FieldConfig::loadMultiple($ids);
+
+    // Clone the field configs, so that we can modify them and change the
+    // target bundle type without manipulating the statically cached entries
+    // in the entity storage;
+    $cloned_field_configs = array_map(static function (FieldConfigInterface $field) use ($bundle) {
+      $cloned = clone $field;
+      $cloned->set('bundle', $bundle);
+      return $cloned;
+    }, $field_configs);
+    foreach ($cloned_field_configs as $field_instance) {
+      $result[$field_instance->getName()] = $field_instance;
+    }
+  }
+  return $result;
+}
+
+/**
+ * Implements hook_entity_view_display_alter().
+ *
+ * There is no way to handle this in the entity type's view builder.
+ */
+function civicrm_entity_entity_view_display_alter(EntityViewDisplayInterface $display, array $context) {
+  $entity_type = \Drupal::entityTypeManager()->getDefinition($context['entity_type']);
+  assert($entity_type !== NULL);
+  if ($entity_type->get('civicrm_entity') && $entity_type->hasKey('bundle')) {
+    $entity_display_repository = \Drupal::service('entity_display.repository');
+    assert($entity_display_repository instanceof EntityDisplayRepositoryInterface);
+    $root_display = $entity_display_repository->getViewDisplay(
+      $entity_type->id(),
+      $entity_type->id()
+    );
+    $display->set('content', $root_display->get('content'));
+    $display->set('hidden', $root_display->get('hidden'));
+  }
 }
 
 /**
--- a/civicrm_entity.module
+++ b/civicrm_entity.module
@@ -19,7 +19,6 @@ use Drupal\Core\Entity\ContentEntityDeleteForm;
 use Drupal\Core\Entity\ContentEntityType;
 use Drupal\Core\Entity\FieldableEntityInterface;
 use Drupal\Core\Field\FieldStorageDefinitionInterface;
-use Drupal\Core\Language\LanguageInterface;
 use Drupal\Core\Render\Element;
 use Drupal\Core\StringTranslation\TranslatableMarkup;
 use Drupal\views\ViewExecutable;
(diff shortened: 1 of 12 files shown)


mglaman:
@KarinG ☝️

KarinG:
Clever! 🆒

Ok we've got tons of them - good thing we organized them a bit 🙂

image

I can't seem to get the CLIN - IWBC to be the same colour though. I'm going to need to have a closer look at this tomorrow. I've cleared caches - but still seeing some of the old colours I had specified (using Taxonomy) 🤔

image

KarinG:
Ok I've created a fresh/new calendar view [independent of my previous colour by taxonomy experiment] -> I've set CLIN - Lunch to green -> I'm still getting default fullcalendarview blue. I've cleared caches. I've disabled Views caching. I keep seeing the blue. 

image

In Settings it says if Taxonomy colour is specified this settings would be ignored:
image

But I don't have Taxonomy colour specified though. This view has no Taxonomy.

jackrabbithanna:
This is interesting. Another entity type that would benefit from this and color coding in calendars is Events with bundles per event type.

mglaman:
@jackrabbithanna I just thought of that today, too! I'm going to update this and it'd allow CiviCRM entity types to specify what field should provide bundle values. I plan on adding Events support in this PR as well since it should be low effort

mglaman:
Okay, I think I figured out the last bit in my last commit.

Screen Shot 2021-01-09 at 1 59 22 PM

KarinG:
Oh... I can't wait to try this! Will do so after this meeting ends.

civicrm_entity.module
```diff
@@ -112,6 +125,33 @@ function civicrm_entity_entity_type_build(array &$entity_types) {
...
+    return !empty($civicrm_entity_info['bundle property']);
+  });
+  foreach ($entity_types_with_bundles as $entity_type_id => $civicrm_entity_info) {
+    $bundles[$entity_type_id] = [];
+    $options = $civicrm_api->getOptions($civicrm_entity_info['civicrm entity name'], $civicrm_entity_info['bundle property']);
+    foreach ($options as $option) {
+      $machine_name = $transliteration->transliterate($option, LanguageInterface::LANGCODE_DEFAULT, '_');
+      $machine_name = mb_strtolower($machine_name);
+      $machine_name = preg_replace('/[^a-z0-9_]+/', '_', $machine_name);
+      $machine_name = preg_replace('/_+/', '_', $machine_name);
```
mglaman: This code exists multiple times

src/CiviEntityStorage.php
```diff
@@ -217,7 +218,23 @@ protected function doLoadMultiple(array $ids = NULL) {
...
+      }
+      $bundle_value = $civicrm_entity[$bundle_property];
+      $options = $this->civicrmApi->getOptions($this->entityType->get('civicrm_entity'), $bundle_property);
+      $bundle = $options[$bundle_value];
+
+      $transliteration = \Drupal::transliteration();
+      $bundle = $transliteration->transliterate($bundle, LanguageInterface::LANGCODE_DEFAULT, '_');
+      $bundle = mb_strtolower($bundle);
+      $bundle = preg_replace('/[^a-z0-9_]+/', '_', $bundle);
+      $bundle = preg_replace('/_+/', '_', $bundle);
```
mglaman: Appears again

src/Plugin/Field/BundleFieldItemList.php
```diff
@@ -0,0 +1,41 @@
...
+    /** @var \Drupal\civicrm_entity\CiviCrmApiInterface $civicrm_api */
+    $civicrm_api = \Drupal::service('civicrm_entity.api');
+    $options = $civicrm_api->getOptions($civicrm_entity_name, $civicrm_bundle_property);
+    $bundle_value = $options[$raw_bundle_value];
+
+    $transliteration = \Drupal::transliteration();
+    $machine_name = $transliteration->transliterate($bundle_value, LanguageInterface::LANGCODE_DEFAULT, '_');
+    $machine_name = mb_strtolower($machine_name);
+    $machine_name = preg_replace('/[^a-z0-9_]+/', '_', $machine_name);
+    $machine_name = preg_replace('/_+/', '_', $machine_name);
```
mglaman: And again

mglaman (commented):
I think there should be a helper function for this bit of code:

```
      $machine_name = $transliteration->transliterate($option, LanguageInterface::LANGCODE_DEFAULT, '_');
      $machine_name = mb_strtolower($machine_name);
      $machine_name = preg_replace('/[^a-z0-9_]+/', '_', $machine_name);
      $machine_name = preg_replace('/_+/', '_', $machine_name);
```

Maybe something which accepts the `$value` and `$transliteration` service. Otherwise, we have the transformation of the value code copied into several places.

mglaman:
@KarinG I _did not_ yet test in conjunction with #264. But given that the list builder displays the bundle field value, now, I think it will work.

KarinG:
I'm getting a conflict:

```
root@25d65da3ebee:/var/www/drupal/web/sites/default/modules/civicrm_entity# git apply 265.diff
error: patch failed: src/Entity/CivicrmEntity.php:2
error: src/Entity/CivicrmEntity.php: patch does not apply
```

Mark could you please merge #264 so that Matt can rebase this and then we've got a clean test of this - on top of the new calculated end time feature?

homotechsual:
The idea here looks awesome to me, this is essentially what we're replicating with CSS classes - this would make life much easier :-)

jackrabbithanna:
Merged #264 so this one can be rebased

mglaman:
Thanks, @jackrabbithanna ! I'll rebase this today

mglaman:
@KarinG okay this is rebased, and I addressed my self-review items for the duplicated code.

KarinG:
Great! Going to be pulling this in shortly.

KarinG:
It's totally working! 🎉 ✅

@mglaman - you're brilliant. 

image

KarinG:
This is move worthy! Take 1: 

https://user-images.githubusercontent.com/5340555/104261646-3b568500-5443-11eb-9c64-9b7dade05730.mov

KarinG:
@mglaman - oh no -> we're hitting an error now trying to Create an Activity using webform:

`CiviCRM_API3_Exception: Missing bundle for entity type civicrm_activity in civicrm_api3() (line 133 of /var/www/drupal/vendor/civicrm/civicrm-core/api/api.php).`

If I `git stash` this PR - then all is well 🤔

homotechsual:
I think we'd need to refactor `webform_civicrm` to account for the bundles as it appears to be using the entity system at some level?

mglaman:
😬  oof. Technically it shouldn't matter, since it's computed. Unless Webform CiviCRM is not specifying an activity type. BUT there could be more gnarly things in the depths of the entity storage.

mglaman:
I got caught up with a busy work week, and I'll be circling up with this on Monday. 

I think the bug is around `civicrm_entity_civicrm_post` or `civicrm_entity_civicrm_pre` which invoke various hooks like `public function civiPreSave` in the entity storage. Since the CiviCRM API is bubbling up a Drupal error message, I have to assume it's due to these hooks.

mglaman:
This is reproducible by even visiting `/civicrm-activity/add`, since there is no default value.

mglaman:
For Webform, I can confirm it is due to `civicrm_entity_civicrm_pre`. So Webform CiviCRM won't need a fix! :)

KarinG:
Oh - is this ready for testing then? I’m at the Endodontist getting a really sore tooth looked after. So should probably get that done first but I can test this afternoon!

mglaman:
@KarinG it is now :) https://github.com/eileenmcnaughton/civicrm_entity/pull/265/commits/f4579990cce3e02aa053766c4862ef1bc2da95d2 fixes the bug Webform was encountering. The previous two commits fixed the user interface (trying to add an Event or Activity would crash)

When adding an activity it will list all bundles (Activity Types) and prefill it on the form (we could even hide it, now.)

Screen Shot 2021-01-18 at 11 41 06 AM

Here is adding an event:

Screen Shot 2021-01-18 at 12 09 05 PM

jackrabbithanna:
I'm putting this into our fork and will let it run on a dev instance for a bit, then if ok a production instance for a bit .. to make sure it is battle tested.

mglaman:
Thanks @jackrabbithanna. I did some shakin' around and Webform CiviCRM tests. That's how I uncovered the UI crash. I think all cases are covered... but as always, there definitely could be hidden gremlins :)

jackrabbithanna:
I see one problem. Trying to add a Drupal field to CiviCRM Activity entity type
The form where you select to add new field or existing, and 
"There was a problem creating field Test Drupal Field: Attempt to create a field 'field_test_drupal_field' without a bundle."
After that error, back to the add field form, but I can't add a field with the same name, because it reports the machine name is in use. 
Adding a field on Contact entity type still works as expected.

mglaman:
Ah, there must be some other place code is needed to "fake" the bundle value. I'll eyeball that.

mglaman:
@jackrabbithanna shoot I realized this might be a breaking change that we can't ship. Custom fields added to Activity and Events _before_ this were on a non-bundle of `civicrm_event`. After this deploy that bundle is invalid. Same with if we ship this, you can't just add a field across all events. Only on a specific bundle.

FieldConfig doesn't support defining fields as a base field. So while this is cool.. I wonder if it's too late and disruptive?

jackrabbithanna:
Hmm. We haven't used fields much, but people could have, although I'd be surprised if it was too many. It is likely there will be 1 or 2. 
Say somebody had created a field on Activity entity type. After this we'd have a bundle per activity type. Can we migrate the field, duplicating it for each bundle? Would there still be only 1 table for the field values? 
Is the LOE for making the UI work per bundle extensive? 
I also feel like if we go with this, we should really think about what entity type could benefit from bundles, and do it for all of them, so we only have to rip the band aid off once. 
Lets give it some more thought, but not necessarily give up.

mglaman:
@jackrabbithanna We definitely could migrate. There is still one database table and field storage – just only one instance of the field config (for a now missing bundle.) So it's as "simple" as an update hook and replicating it for the new bundles and then removing the old field config.

I'm not sure the LOE for making it work, I would need _some_ initial dig in. I know the Field and Field UI code is something that isn't _bad_ but it hasn't been an area folks worked on (it worked and works for core's standard needs.)

I'll do some initial research on the level of effort. I'm pretty sure there is a way we can make it work fairly seamlessly and preserve existing functionality.

jackrabbithanna:
@mglaman Ok, thank you!
I feel like we should make an effort to have an upgrade process, and not burn those people that are using this feature already. We did get feedback from at least one company using this feature. There is no rush, I'm thinking to release beta2 with a lot of updates and fixes, and this can be for a next release after that.

mglaman:
So adding some notes for myself here.

This is from `\Drupal\field_ui\Routing\RouteSubscriber::alterRoutes`. If there isn't a `{bundle}` parameter in the `field_ui_base_route` it assumes the bundle is the entity type ID.

```
        $defaults = [
          'entity_type_id' => $entity_type_id,
        ];
        // If the entity type has no bundles and it doesn't use {bundle} in its
        // admin path, use the entity type.
        if (strpos($path, '{bundle}') === FALSE) {
          $defaults['bundle'] = !$entity_type->hasKey('bundle') ? $entity_type_id : '';
        }
```

In `\Drupal\field_ui\Form\FieldStorageAddForm::submitForm` it sets the field values as

```
      $field_values = [
        'field_name' => $values['field_name'],
        'entity_type' => $this->entityTypeId,
        'bundle' => $this->bundle,
        'label' => $values['label'],
        // Field translatability should be explicitly enabled by the users.
        'translatable' => FALSE,
      ];
```

So I think we could alter `entity.field_config.{$entity_type_id}_field_edit_form` for CiviCRM entities to use a custom form that ensures there is a field across each bundle. But the field config listing would prove to be another challenge.

`\Drupal\field_ui\FieldConfigListBuilder::load` loads by `\Drupal\Core\Entity\EntityFieldManagerInterface::getFieldDefinitions` with the entity type ID and bundle. _However_ we know the bundles all have the same fields. So we could override this as well to load all across each bundle or just the first one since we know it is replicated.

So my thoughts:

Route altering for CiviCRM Entities:

* `"entity.{$entity_type_id}.field_ui_fields"`: Extend `FieldConfigListController` so that `load` includes all bundles, not the bundle from the route (since it does not exist.)
* `"field_ui.field_storage_config_add_$entity_type_id"`: Extend `FieldStorageAddForm` to create a FieldConfig for each bundle on submit.

But then I realized we're going to have more cascading overrides when dealing with the "Manage form display" and "Manage display".

We might need to add a "types" listing next to `"entity.$entity_type_id.collection"` which each Field UI builds off of. We'll want to copy fields onto each bundle still. But this would allow custom form displays by type and display.

One question is: is this too much? Data doesn't differ much between the Event Type or Activity Type. This is where Drupalisms could cause problems.

Again, however, if we want to support bundles _but not_ individual bundle view display or form display, we can create a new Field UI layer. At this point either way is probably the same effort. It is just needing to know "what is the best approach"

mglaman:
I'm leaning towards "building a new Field UI" just for CiviCRM Entities. That way we don't need `/admin/structure/civicrm-entity/civicrm-activity/{bundle}/fields` and such. We just make a new layer for

* `/admin/structure/civicrm-entity/civicrm-activity/fields`
* `/admin/structure/civicrm-entity/civicrm-activity/form-display`
* `/admin/structure/civicrm-entity/civicrm-activity/display`

Behind the scenes, we'll handle if the CiviCRM Entity type supports bundles or not. If it doesn't, current functionality. If it does, we ensure things are copied across all bundles as Drupal expects.

civicrm_entity.module
```diff
@@ -109,10 +115,108 @@ function civicrm_entity_entity_type_build(array &$entity_types) {
...
+    return !empty($civicrm_entity_info['bundle property']);
+  });
+  foreach ($entity_types_with_bundles as $entity_type_id => $civicrm_entity_info) {
+    // We keep a bundle that is the same as the entity type ID. This allows us
+    // to create fields as if this entity has no bundles.
+    $bundles[$entity_type_id] = [
+      $entity_type_id => [
+        'label' => $civicrm_entity_info['civicrm entity label'],
+      ],
+    ];
```
mglaman, line 157: Here is a trick! So we keep the existing bundle as entity type ID along with all of the extra options. Technically nothing will be this value. But it preserves backward compatibility _and_ helps make the user interface work.

civicrm_entity.module
```diff
@@ -109,10 +115,108 @@ function civicrm_entity_entity_type_build(array &$entity_types) {
...
+    $cloned_field_configs = array_map(static function (FieldConfigInterface $field) use ($bundle) {
+      $cloned = clone $field;
+      $cloned->set('bundle', $bundle);
+      return $cloned;
+    }, $field_configs);
+    foreach ($cloned_field_configs as $field_instance) {
+      $result[$field_instance->getName()] = $field_instance;
+    }
+  }
+  return $result;
```
mglaman, line 199: We copy the logic in `field.module` to define bundle fields. The "trick" is that we load field configs for the entity type. We make clone of the object – so we don't pollute existing references – and change the bundle. This acts almost like a deriver and makes derivatives of each field config for each bundle.

civicrm_entity.module
```diff
@@ -109,10 +115,108 @@ function civicrm_entity_entity_type_build(array &$entity_types) {
...
+    $entity_display_repository = \Drupal::service('entity_display.repository');
+    assert($entity_display_repository instanceof EntityDisplayRepositoryInterface);
+    $root_display = $entity_display_repository->getViewDisplay(
+      $entity_type->id(),
+      $entity_type->id()
+    );
+    $display->set('content', $root_display->get('content'));
+    $display->set('hidden', $root_display->get('hidden'));
+  }
+}
```
mglaman, line 220: Ugh, unfortunately, there is no way to say "no, don't use civicrm_activity.meeting.default, use civicrm_activity.civicrm_activity." So we have to do this, here.

civicrm_entity.services.yml
```diff
@@ -11,6 +11,13 @@ services:
     arguments: ['@civicrm_entity.api']
     tags:
       - { name: backend_overridable }
+
+  civicrm_entity.field_ui_route_subscriber:
+    class: Drupal\civicrm_entity\Routing\RouteSubscriber
+    arguments: [ '@entity_type.manager' ]
+    tags:
+      - { name: event_subscriber }
```
mglaman, line 19: Fixes the bundle default route param.

mglaman (commented):
I think this fixes the Field UI and prevents needing any kind of upgrade requirements. We'd want someone to test this branch out with their code who has existing fields.

I'll reinstall my local environment without this branch and setup some fields. Then I'll switch over to this branch. There should be no need for upgrades. Just a cache rebuild for new hook discovery – so it would be worth adding that as a post update. I'll work on that as part of my testing.

civicrm_entity.module
```diff
@@ -125,8 +125,10 @@ function civicrm_entity_entity_type_build(array &$entity_types) {
     if (!empty($civicrm_entity_info['bundle property'])) {
       $entity_type_info['entity_keys']['bundle'] = 'bundle';
       $entity_type_info['civicrm_bundle_property'] = $civicrm_entity_info['bundle property'];
-      $entity_type_info['links']['add-page'] = $entity_type_info['links']['add-form'];
-      $entity_type_info['links']['add-form'] = sprintf('%s/{%s}', $entity_type_info['links']['add-form'], $entity_type_info['entity_keys']['bundle']);
+      if (isset($entity_type_info['links'])) {
+        $entity_type_info['links']['add-page'] = $entity_type_info['links']['add-form'];
+        $entity_type_info['links']['add-form'] = sprintf('%s/{%s}', $entity_type_info['links']['add-form'], $entity_type_info['entity_keys']['bundle']);
+      }
```
mglaman: I didn't notice this before, but it would set null routes if the entity wasn't exposed to the UI – like Activity. This fixes if the entity is not yet exposed to the UI

civicrm_entity.post_update.php
```diff
@@ -6,7 +6,7 @@
  */
 
 /**
- * Implements hook_post_update_NAME().
+ * Enable bundle support for CiviCRM Entity entity types.
```
mglaman: Fix title.

mglaman (commented):
Fixed the upgrade

KarinG:
Testing this right now!

```
git pull
curl https://patch-diff.githubusercontent.com/raw/eileenmcnaughton/civicrm_entity/pull/265.diff --output 265.diff
git apply 265.diff
sudo -u www-data drush updatedb
Do you wish to run the specified pending updates? (yes/no) [yes]:
 > yes
>  [notice] Update started: civicrm_entity_post_update_entity_bundles
>  [notice] Update completed: civicrm_entity_post_update_entity_bundles
 [success] Finished performing updates.
```
for good measure:
```
sudo -u www-data drush cr
```

tested 🎉
- Calendar colour config ✅
- Create activity ✅
- Upgrade -> I can see my pre-existing Taxonomy Entity Reference field on civicrm-entity/civicrm-activity/fields ✅

Mark -> this is ready for you!

KarinG:
Must add a screenshot! 

image

KarinG:
@jackrabbithanna - to give this PR some exposure ahead of a release perhaps could it be merged so that people installing from git will start using it?

jackrabbithanna:
I haven't had time for another round of testing yet, but I'll merge it as long as if something comes up, you all can tweak it. 
Merging #265

mglaman:
Ping me if anything goes awry for users

codebymikey:
This PR appears to introduce a couple breaking change in behaviour on the Event entity type (and I'm assuming other bundle entities), similar to the one [described here](https://github.com/eileenmcnaughton/civicrm_entity/pull/265#issuecomment-762401719).

Visiting `/civicrm-event/add`  shows a list of Event types, however the only fully functional one is the "Event".
The other bundle types still require the "Event Type" field to be set, which is confusing and can potentially break the site.
The custom form display settings only apply on the bundles rather than the core `/civicrm-event/add/civicrm_event`.

The following warning is also shown on any non-Event bundles:
```
Notice: Undefined index: event_type_id in Drupal\civicrm_entity\Entity\CivicrmEntity::preCreate() (line 87 of modules/contrib/civicrm_entity/src/Entity/CivicrmEntity.php).
Drupal\civicrm_entity\Entity\CivicrmEntity::preCreate(Object, Array) (Line: 209)
```

The line in question is this:
https://github.com/eileenmcnaughton/civicrm_entity/blob/4e6ec710f2ac88cc7510354a8af293f86e00bf3b/src/Entity/CivicrmEntity.php#L83-L88

KarinG:
You ran updatedb?

Show me exactly what the breaking changes are: 

screenshots; steps to reproduce; what happens; what’s broken; -> then I can circle back to this (with MattG if needed).

Example URL of where you’re seeing the notice on?

D8 or D9?

homotechsual:
I haven't been able to reproduce this on Drupal 9.1.x on our sites.

Going to need more information to reproduce this and get it fixed :-)

codebymikey:
That's interesting to hear, I'm pretty sure `updatedb` was ran following the update.
I'll set some time aside to test on a clean install and see if I can replicate the behaviour/issues then - hopefully it's just a misconfiguration somewhere rather than a bug.

Thanks for looking into it, and providing a second set of feedback on it.
I think it's most likely related to the entity/field definition changes, [and the update hooks](https://www.drupal.org/node/3034742) for them not being ran for some reason:

image

I'll try out the clean install as well as if [`devel_entity_updates`](https://www.drupal.org/project/devel_entity_updates) also fixes the existing behaviour.

jackrabbithanna:
We had fixed the "mixmatched entity field definition" issue in beta2, are you using that? 
It should fix that up on cache clears.

codebymikey:
Apologies for the red-herring, the "mismatched entity" issue came back because we ended up downgrading back to a version before this commit.

I've attached a video showcasing the behaviour (on the latest dev):
<video src="https://user-images.githubusercontent.com/9484406/112673441-a0bfec80-8e5c-11eb-9eb6-9641de08d547.mp4"></video>

- The event bundles show up on the "Add civicrm event" page (`/civicrm-event/add`).
- On clicking an Event bundle, it fails to select the "Event Type" field because the IDs on the select box are numeric, whereas the bundle property being passed in is a human-friendly string, causing the undefined index warnings.
- I believe that if you're on a Bundle page, you should no longer have the option to switch the bundle type (since it's confusing and there might be an edge case where a different bundle has additional required fields/conditions which the original doesn't) - the field should be set "internally".

KarinG:
I just tried to reproduce this (same steps as in your screencast) -> and I do not get any notice(s) in the dblog. I am able to successfully see it through and create a Conference.  

I'm on 8.9.13 + latest civicrm entity - master. Did you run `drush updatedb` after checking out the latest?

jackrabbithanna:
I think codebymikey makes a good point about:
"I believe that if you're on a Bundle page, you should no longer have the option to switch the bundle type (since it's confusing and there might be an edge case where a different bundle has additional required fields/conditions which the original doesn't) - the field should be set "internally"."

There's a few things to consider / fix here. 
1) Perhaps we should automatically set the "bundle" field .. e.g "Activity Type" for add/edit of activities, by the chosen bundle 
2) Should we disable this field on "add" .. and/or on "edit" ? I think yes for the "add" forms .. Now for "edit" .. the question is do we disable the field OR do we allow changing of activity type? If activity type changes, then the bundle should change. 
3) I do see notices that need fixing . 3 notices appear in the log for the activity "add" form:
Notice: Undefined index: activity_type_id in Drupal\civicrm_entity\Entity\CivicrmEntity::preCreate() (line 87 of /var/www/web/demo8.skvare.com/web/modules/contrib/civicrm_entity/src/Entity/CivicrmEntity.php) 

otice: Undefined index: in Drupal\civicrm_entity\Entity\CivicrmEntity::preCreate() (line 90 of /var/www/web/demo8.skvare.com/web/modules/contrib/civicrm_entity/src/Entity/CivicrmEntity.php) 

Notice: Undefined index: in Drupal\civicrm_entity\Plugin\Field\BundleFieldItemList->setValue() (line 60 of /var/www/web/demo8.skvare.com/web/modules/contrib/civicrm_entity/src/Plugin/Field/BundleFieldItemList.php) 

4) this part of the statement "there might be an edge case where a different bundle has additional required fields/conditions which the original doesn't"

This could be out of scope for this issue. No version of CE has ever ONLY put the custom fields that are set for the activity type / event type on the form. This is a feature improvement that would be great to have one day, but treated separately. 

As it is now, we've done things where we have hook_form_alter() implementations to unset or hide custom fields based on activity type etc.. 

@KarinG @mglaman Please let me know your thoughts. 
Thank you!

jackrabbithanna:
@KarinG 
We had been planning to push another release to drupal.org ahead of our upcoming webinar, but cannot include this feature yet because of the log messages, and it needs a little more thought for how to handle the add/edit forms. 

Please let me know if you can provide updates for 3) and discuss ideas/options/improvements for 1) and 2). 

Depending on that we can hold on a release, or revert this so we can push out all other updates in a new release. 
We're supportive of getting this functionality into an official release when it is finalized.

Thank you!

jackrabbithanna:
This also breaks usage of Layout Builder on the "Manage Display" area

genaskvare:
@puresyntax71 - please review the issues found above and see about collaborating with @KarinG and @mglaman to support getting the items addressed and tested. Let's just focus on a clean release. If that means taking more time and releasing later, that's time worth taking! I don't want this reverted. Thanks.

KarinG:
Since this PR is Merged I'll open up an ad-hoc issue (here in Github) re: bundles ad-hoc.

KarinG:
Ok -> Issue we can work off is here: https://github.com/eileenmcnaughton/civicrm_entity/issues/285

PR #266, opened 2021-01-13
Fixed variable $columns overwrite into civicrm_entity_views_query_alt…

GValFr35:
…er()

+ Improve code logic


GValFr35:
Fix PHP warnings when calling array_key_exists () on string.
This is due to an array ($columns) which is replaced by a string during the previous loop.

PR #267, opened 2021-01-13
Fix html type RichTextEditor for custom fields

GValFr35:
Civicrm custom fields using RichTextEditor are not displayed in HTML format in Drupal. The TextareaWidget with the configurated text format does not apply.
This is due to a difference in array keys returned by the getfields apiv3 between primary fields and custom fields.
$civicrm_field['html_type'] instead of $civicrm_field['html']['type']

### DIFF
--- a/src/Entity/FieldDefinitionProvider.php
+++ b/src/Entity/FieldDefinitionProvider.php
@@ -256,7 +256,13 @@ class FieldDefinitionProvider implements FieldDefinitionProviderInterface {
    *   The base field definition.
    */
   protected function getTextDefinition(array $civicrm_field) {
-    if (!empty($civicrm_field['html']) && $civicrm_field['html']['type'] == 'RichTextEditor') {
+
+    if (!empty($civicrm_field['html_type']) && $civicrm_field['html_type'] == 'RichTextEditor') {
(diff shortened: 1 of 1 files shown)


jackrabbithanna:
The API is inconsistent in the property metadata returned by 'getfields' api action. 
Most of the time, you see $civicrm_field['html']['type']
I see there are a few cases where $civicrm_field['html_type'] exists and has a value. 

I think it should be something like this:
```
if ((!empty($civicrm_field['html']) && $civicrm_field['html']['type'] == 'RichTextEditor') ||
  (!empty($civicrm_field['html_type']) && $civicrm_field['html_type'] == 'RichTextEditor')) {
```
Would you like to update your PR with this?

jackrabbithanna:
Actually this would be better:
```
if ((!empty($civicrm_field['html']['type']) && $civicrm_field['html']['type'] == 'RichTextEditor') ||
  (!empty($civicrm_field['html_type']) && $civicrm_field['html_type'] == 'RichTextEditor')) {
```

jackrabbithanna:
Closing in favor of: https://github.com/eileenmcnaughton/civicrm_entity/pull/294

PR #280, opened 2021-03-09
Update civicrm_entity_price_set_field.event_registration.inc

Michael-Labriola:
When available, use payment processor title instead of name.  I believe the title field was created to avoid exposing internal names on public forms.


jackrabbithanna:
Yep, I think we should be using titles instead of names to display to the user. I'm good with this. Merging. Will be pushing to D.O very soon in a new release.

PR #281, opened 2021-03-15
Add views relationships to tags.

puresyntax71:


jackrabbithanna:
Adds Views relationship Contacts -> Tags, Tags -> Contacts
Bridges through Entity Tag entity type
To use, start with list of contacts, add Views relationship "CiviCRM contact related to the CiviCRM tag"
You can now filter the list of contacts by any tag property, and list will show contacts which have that tag(s)

PR #282, opened 2021-03-15
Set proper required configuration.

puresyntax71:


jackrabbithanna:
For the Drupal based add/edit forms, if a CiviCRM custom field is marked as required in it's settings, then the form will require the field.

PR #283, opened 2021-03-19
load multiple records via a single API call, not one call per

MegaphoneJon:
`Drupal\civicrm_entity\CiviEntityStorage::doLoadMultiple()` takes a set of record IDs and loads the data from Civi with a separate API call for each record.  I was able to get impressive performance boosts by combining them into a single API call - about 30% from a cold start, and 50% when cached.

I used this command to get timing in milliseconds for rendering the view:
```shell
BEGIN=$(($(date +%s%N)/1000000)) && drush vex member_directory 2&>1 > /dev/null && expr $(($(date +%s%N)/1000000)) - $BEGIN
```
I ran it twice each with cache cleared, and four times each after a no-cache run to get cached values.

Before fix no cache: 5219, 5194
Before fix cached: 2801, 2889, 2926, 2948

After fix no cache: 3632, 3738
After fix caches: 1467, 1346, 1352, 1321

### DIFF
--- a/src/CiviEntityStorage.php
+++ b/src/CiviEntityStorage.php
@@ -158,12 +158,10 @@ class CiviEntityStorage extends SqlContentEntityStorage {
   protected function doLoadMultiple(array $ids = NULL) {
     $entities = [];
     if ($ids === NULL) {
-      $civicrm_entities = $this->getCiviCrmApi()->get($this->entityType->get('civicrm_entity'));
-      foreach ($civicrm_entities as $civicrm_entity) {
-        $civicrm_entity = reset($civicrm_entity);
-        $entity = $this->prepareLoadedEntity($civicrm_entity);
-        $entities[$entity->id()] = $entity;
-      }
+      $civicrm_entity = $this->getCiviCrmApi()->get($this->entityType->get('civicrm_entity'));
+      $civicrm_entity = reset($civicrm_entity);
+      $entity = $this->prepareLoadedEntity($civicrm_entity);
+      $entities[$entity->id()] = $entity;
     }
 
     // get all the fields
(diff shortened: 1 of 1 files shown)


MegaphoneJon:
2 notes:
* This PR looks bigger than it is because of indentation.  Use "hide whitespace changes" when viewing the diff and you'll get a better picture.
* I'm still having issues RAM-wise because we're loading the contents of every custom field, and these contacts have a couple hundred of them, many of them long text fields.  The abstraction may render it impossible (or at least unclear to a Drupal neophyte like me) how to get `doLoadMultiple` to only request fields needed for the View.

jackrabbithanna:
Yes! I really like this idea here. We will be reviewing / testing this. Thanks Jon!
You said:
"I'm still having issues RAM-wise because we're loading the contents of every custom field, and these contacts have a couple hundred of them, many of them long text fields. The abstraction may render it impossible (or at least unclear to a Drupal neophyte like me) how to get doLoadMultiple to only request fields needed for the View."

We've had similar problems, either with RAM, or just horrible performance with datasets like with 1Million contact records. 
So you're thinking loading the entity, but only populating values for those fields you need for the View .. hmm .. yep. 

Unfortunately in D8/D9 Views does an entity load .. not sure if it's possible in Drupal to do what you are thinking. 

One possible alternative is to think about using Search API .. with the "Database search" or even Apache Solr .. so that you can index only what you need and base a View on that.

MegaphoneJon:
@jackrabbithanna I [opened an issue](https://www.drupal.org/project/civicrm_entity/issues/3204626) on d.o.  It should be fine to do an entity load so long as we don't pass all the fields to `initFieldValues()`.

jackrabbithanna:
I'm testing this PR and it seems fine so far. 
wrt only loading what fields are necessary for the View:

doLoadMultiple() calls prepareLoadedEntity() which calls initFieldValues() 
doLoadMultiple() is only passed ids, the same as the [parent class](https://api.drupal.org/api/drupal/core%21lib%21Drupal%21Core%21Entity%21Sql%21SqlContentEntityStorage.php/function/SqlContentEntityStorage%3A%3AdoLoadMultiple/8.9.x) has.

The problem is how to get Views to load only those fields that are needed for that View, which as far as I know other entity type like Nodes or Users behave the same way with Views as with CiviCRM Contact. 

At the same time, much other functionality will depend on all fields being loaded by doLoadMultiple()

I do not know how to achieve what you are looking for currently. We will see if others provide feedback or we learn something in the future that can help. 

Thinking outside the box, it is possible to create a MySQL View, and then make custom Views integration for that. Make Drupal views see your MySQL view and offer it as an option for a base table. Could be another way to get what you need in the now, albeit with a lot of elbow grease.

jackrabbithanna:
@KarinG requested that we wait to merge this until she and/or others have had a chance to test it more. 

We've been using this in demo/dev sites with success, and will be moving to production soon. No issues so far, but I like the idea of "better safe than sorry"

KarinG:
To clarify -> @jackrabbithanna - you said you were preparing for a D8 Release and wanted to merge this to get it into the Release. That felt rushed based on what you had documented prior to yesterday "I'm testing this PR and it seems fine so far."

jackrabbithanna:
Just released 8.x-3.0-beta3, merging this now for github repo users

PR #284, opened 2021-04-06
Add uri callback to civicrm_contact entity type

jitendrapurohit:
I think it need to be discussed but it fixes the problem raised in https://github.com/eileenmcnaughton/civicrm_entity/issues/275

Links to the `entity_type` option is available on almost all entities?

Eg contact fields display 
image

Contribution fields eg source etc displays -

image

Currently, all of them fail to link to the entity. This PR fixes the problem for contact fields. 

For others, probably we should just replace the `if()` with `switch()` and handle it for all entity types?

### DIFF
--- a/civicrm_entity.module
+++ b/civicrm_entity.module
@@ -24,6 +24,7 @@ use Drupal\Core\Entity\FieldableEntityInterface;
 use Drupal\Core\Field\FieldStorageDefinitionInterface;
 use Drupal\Core\Render\Element;
 use Drupal\Core\StringTranslation\TranslatableMarkup;
+use Drupal\Core\Url;
 use Drupal\field\Entity\FieldConfig;
 use Drupal\field\FieldConfigInterface;
 use Drupal\views\ViewExecutable;
@@ -79,6 +80,7 @@ function civicrm_entity_entity_type_build(array &$entity_types) {
         'label' => $civicrm_entity_info['label property'],
       ],
       'base_table' => $entity_type_id,
+      'uri_callback' => 'entity_uri',
       'admin_permission' => 'administer civicrm entity',
       'permission_granularity' => 'entity_type',
       'handlers' => [
@@ -133,6 +135,18 @@ function civicrm_entity_entity_type_build(array &$entity_types) {
 
     $entity_types[$entity_type_id] = new ContentEntityType($entity_type_info);
   }
+}
+
+/**
+ * Entity URI callback used in hook_entity_type_build().
+ */
+function entity_uri($entity) {
+  if ($entity->getEntityTypeId() == 'civicrm_contact') {
+    return Url::fromUri('internal:/civicrm/contact/view', [
+      'query' => ['cid' => $entity->id(), 'reset' => 1],
+    ]);
+  }
+  return NULL;
 }
 
 /**


jackrabbithanna:
I believe the intent here is to link to the Drupal version of the entity view page..so for contacts:
/civicrm-contact/[contact-id]

Then a pattern can be used to output the links .. so entity type = civicrm_contact, link to /civicrm-contact/[id] if entity type = civicrm_activity, link to /civicrm-activity/[id]

I was not able to replicate the original issue yet, so was either fixed in our fork, or something else in play which makes this work. 
Will be looking into it.

KarinG:
Historically this link has been to a `/civicrm` page (as @jitendrapurohit has now implemented). For `/civicrm-contact/[contact-id]` -> to work one needs to have checked off the checkbox in the CE Settings config right? Not everyone may want to expose Contacts like that? I suppose there can always be two links -> one to the contact summary page in CiviCRM itself and one to the `civicrm-contact/[contact-id]`?

jackrabbithanna:
When I say "original intent" .. just the Entity API and entities in general link to their Drupal view page traditionally. 
We use the current feature to link to the drupal version of the page. 
Like if you "Link to content" with the Node's title field, you go to the node view page. 

So people coming from a Drupal dev/site builder perspective, may like it going to /civicrm-contact/[id] .. so i have an interest in keeping CE as Drupal-ly as possibly

I completely agree it would be great to have a Views field, or an option to link to the civicrm admin contact page. 

Many entity types it doesn't make sense or need to go to a civicrm admin page either .. maybe just some like Contact, Event .. What else?

Good point about the link not pointing to a page that exists if "CiviCRM Contact" is not checked on the settings page. That may be why the original reporter has the error. 

So .. how to have the cake and eat it too is the question. Let's think about "both" options. 

Maybe for those entity types we decide this is important, we try to add an additional checkbox?

KarinG:
In D7 / CiviCRM Views integration [even with CiviCRM Entity - incl Contact enabled] -> this checkbox -> links the field to its `/civicrm` contact record. 

image

That's how everyone who has built D7 / CiviCRM Views before and are now getting in D8 / CiviCRM Views will be expecting.

jackrabbithanna:
Another thing to think about, is that setting the entity_uri like this MR does, has more effects than just the checkbox in Views. 

That would become the default link, one example is for contacts referenced in Entity Reference fields when using the "Label" field formatter. 

You're not thinking about the total picture yet, all the use cases, and the consequences. 
This needs more thought. For the "entity reference" use case this PR will not work. 

I have to think about all the consequences for all the Drupal integration, not just Views, and not just what was in D7.

It would be better if a custom or additional Views field, "Link to CiviCRM admin contact page" .. or an additional checkbox, or a field formatter to accomplish this bit of Views functionality. Some way that provides the functionality for Views, but doesn't affect other things.

jackrabbithanna:
Discovered that the error in the issue referenced can occur by not having "CiviCRM Contact" enabled as an entity type on the setting page. 
Site builders who want to link to civicrm admin page can do:

Add Contact ID to the View (can exclude from display)
Add display name (or any field) 
In display name field settings goto "Rewrite Results" fieldset
Check the "Output this field as a custom link"
For link path enter: /civicrm/contact/view?reset=1&cid={{ id }}

I'm happy to discuss methods to alleviate the UI/UX issue with this checkbox. 

I'm also happy to discuss adding an easier way to link to CiviCRM admin contact page. 

I cannot merge this PR as it is, due to it breaking other functionality's expected behavior.

Closing for now, but happy to re-open and/or work on these things with anybody, with a fresh approach.

PR #286, opened 2021-04-09
Fixes #285

### LINKED ISSUE #285: Bundles ad-hoc
Issues to be addressed:

**1. [this will no longer be an issue after 2] - Notices -> activity "add" form:** [KarinG: I have not been able to reproduce these but will try again and circle back here]
Notice: Undefined index: activity_type_id in Drupal\civicrm_entity\Entity\CivicrmEntity::preCreate() (line 87 of /var/www/web/demo8.skvare.com/web/modules/contrib/civicrm_entity/src/Entity/CivicrmEntity.php)

Notice: Undefined index: in Drupal\civicrm_entity\Entity\CivicrmEntity::preCreate() (line 90 of /var/www/web/demo8.skvare.com/web/modules/contrib/civicrm_entity/src/Entity/CivicrmEntity.php)

Notice: Undefined index: in Drupal\civicrm_entity\Plugin\Field\BundleFieldItemList->setValue() (line 60 of /var/www/web/demo8.skvare.com/web/modules/contrib/civicrm_entity/src/Plugin/Field/BundleFieldItemList.php)

**2. Bundles are showing up in places where they don't really need to be showing and that is confusing.** Also some larger clients can easily have 500+ activity types -> which results in lots of pages here. Easiest solution may be to not display the bundles where they are not needed. @mglaman thought he had already done this. He'll check into this -> 

image

**3. Layout Builder issue - Manage display**  [KarinG: I'll try and reproduce - we don't use Layout Builder but obviously don't want to break anything].

cc: @puresyntax71 @mglaman @genaskvare @jackrabbithanna

puresyntax71:


src/Routing/RouteSubscriber.php
```diff
@@ -69,12 +69,15 @@ protected function alterRoutes(RouteCollection $collection) {
         "entity.entity_view_display.{$entity_type_id}.view_mode" => [
           'bundle' => $entity_type_id,
         ],
+        "layout_builder.defaults.$entity_type_id.view" => [
+          'bundle' => $entity_type_id,
+        ],
```
KarinG, line 74: @puresyntax71 -> I think it may need `{` and `}` -> `layout_builder.defaults.{$entity_type_id}.view`?

KarinG: But that's not solving the view issue :-(

src/Routing/RouteSubscriber.php
```diff
@@ -69,12 +69,15 @@ protected function alterRoutes(RouteCollection $collection) {
...
       ];
       foreach ($field_ui_routes as $route_name => $defaults) {
-        $route = $collection->get($route_name);
-        assert($route !== NULL);
-        foreach ($defaults as $name => $default) {
-          $route->setDefault($name, $default);
+        if ($route = $collection->get($route_name)) {
+          foreach ($defaults as $name => $default) {
+            $route->setDefault($name, $default);
+          }
```
mglaman, line 80: I'm not sure I see any consequential changes here?

mglaman: I see, it's doing null check, which shouldn't truly return null unless the Field UI module is not installed. In that case, we shouldn't run this at all.

mglaman:
See https://github.com/eileenmcnaughton/civicrm_entity/pull/289, this PR didn't address the `add_page` and `add_form` changes completely.

PR #289, opened 2021-04-11
Remove add_page, simplify adding bundle supported entities, Layout Builder fixess

mglaman:
Addresses items 1 & 2 from #285. There is no longer an `add_page` which lists all of the bundles.

### DIFF
--- a/civicrm_entity.module
+++ b/civicrm_entity.module
@@ -126,7 +126,9 @@ function civicrm_entity_entity_type_build(array &$entity_types) {
       $entity_type_info['entity_keys']['bundle'] = 'bundle';
       $entity_type_info['civicrm_bundle_property'] = $civicrm_entity_info['bundle property'];
       if (isset($entity_type_info['links'])) {
-        $entity_type_info['links']['add-page'] = $entity_type_info['links']['add-form'];
+        // For entities with bundles that are exposed, add the `bundle` key to
+        // the add-form route. In CiviCrmEntityRouteProvider::getAddFormRoute
+        // we default the value, so that it isn't actually required in the URL.
         $entity_type_info['links']['add-form'] = sprintf('%s/{%s}', $entity_type_info['links']['add-form'], $entity_type_info['entity_keys']['bundle']);
       }
     }
--- a/civicrm_entity.module
+++ b/civicrm_entity.module
@@ -26,6 +26,7 @@ use Drupal\Core\Render\Element;
 use Drupal\Core\StringTranslation\TranslatableMarkup;
 use Drupal\field\Entity\FieldConfig;
 use Drupal\field\FieldConfigInterface;
+use Drupal\layout_builder\Entity\LayoutBuilderEntityViewDisplay;
 use Drupal\views\ViewExecutable;
 use Drupal\views\Plugin\views\query\QueryPluginBase;
 
@@ -218,6 +219,13 @@ function civicrm_entity_entity_view_display_alter(EntityViewDisplayInterface $di
     );
     $display->set('content', $root_display->get('content'));
     $display->set('hidden', $root_display->get('hidden'));
+
+    if ($root_display instanceof LayoutBuilderEntityViewDisplay) {
+      $layout_builder_settings = $root_display->getThirdPartySettings('layout_builder');
+      foreach ($layout_builder_settings as $setting_key => $setting) {
+        $display->setThirdPartySetting('layout_builder', $setting_key, $setting);
+      }
+    }
   }
 }
 
--- a/src/Routing/RouteSubscriber.php
+++ b/src/Routing/RouteSubscriber.php
@@ -48,6 +48,8 @@ final class RouteSubscriber extends RouteSubscriberBase {
     if (!$this->moduleHandler->moduleExists('field_ui')) {
       return;
     }
+
+    $has_layout_builder = $this->moduleHandler->moduleExists('layout_builder');
     foreach ($this->entityTypeManager->getDefinitions() as $entity_type_id => $entity_type) {
       if (!$entity_type->get('civicrm_entity_ui_exposed')) {
         continue;
@@ -85,6 +87,20 @@ final class RouteSubscriber extends RouteSubscriberBase {
           'bundle' => $entity_type_id,
         ],
       ];
+
+      if ($has_layout_builder) {
+        // @todo we should iterate over the section storage definitions.
+        //   that means we'd need to conditionally inject the manage service.
+        // @see \Drupal\layout_builder\Plugin\SectionStorage\DefaultsSectionStorage::buildRoutes
+        $field_ui_routes["layout_builder.defaults.$entity_type_id.view"] = [
+          'bundle' => $entity_type_id,
+        ];
+        // @see \Drupal\layout_builder\Plugin\SectionStorage\OverridesSectionStorage::buildRoutes
+        $field_ui_routes["layout_builder.overrides.$entity_type_id.view"] = [
+          'bundle' => $entity_type_id,
+        ];
+      }
+
       foreach ($field_ui_routes as $route_name => $defaults) {
         $route = $collection->get($route_name);
         assert($route !== NULL);
--- a/src/Routing/RouteSubscriber.php
+++ b/src/Routing/RouteSubscriber.php
@@ -3,6 +3,7 @@
 namespace Drupal\civicrm_entity\Routing;
 
 use Drupal\Core\Entity\EntityTypeManagerInterface;
+use Drupal\Core\Extension\ModuleHandlerInterface;
 use Drupal\Core\Routing\RouteSubscriberBase;
 use Drupal\Core\Routing\RoutingEvents;
 use Symfony\Component\Routing\RouteCollection;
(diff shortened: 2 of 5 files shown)


KarinG:
Thanks 🎉 Matt. I'll be testing this re: 1 and 2 now!

KarinG:
Ok -> I rolled back to `f5eaa5c` & applied this PR 289 -> tested! 

Confirming 2 is resolved ✅

Re: 1 -> I do see one notice now -> on the `/civicrm-activity/add` -> [I only filled in date and time; activity type: meeting]; PS - you would not be seeing this in `/civicrm-event/add` testing -> this field is specific to activities. So I submitted `/civicrm-activity/add` without `source_contact_id` filled in. In CiviCRM native this would default to the logged in user. Making it required would work too. Or for now - just a check to ensure there is no notice - would work for now to get this bundle work PR over the finish line.

```
Referrer | http://d8civicrm.local/civicrm-activity/add
Notice: Undefined index: source_contact_id in Drupal\civicrm_entity\Entity\CivicrmEntity->validate() (line 174 of /Applications/MAMP/htdocs/d8civicrm.local/web/modules/contrib/civicrm_entity/src/Entity/CivicrmEntity.php)

```

KarinG:
PS - re: `source_contact_id` -> on `/civicrm-activity/form-display` -> I found `Activity Source Contact` in the list of Disabled fields -> Enabling it -> It is on the form now and indeed (correctly) is required  -> problem solved. No more notice :-)

image

So 1. and 2. are 👍 Just the Form Builder issues are left.

mglaman:
Oh! Okay, so we need to update the list of default form displays, then. Weird that this was not a problem before.

mglaman:
Let's open up the `source_contact_id` warning as its own issue that `source_contact_id` should prepopulate to the currently logged-in user since it is a required field. If we can reproduce this outside of this PR, then I don't think we should fix it here.

KarinG:
Agreed 👍

civicrm_entity.module
```diff
@@ -216,6 +219,13 @@ function civicrm_entity_entity_view_display_alter(EntityViewDisplayInterface $di
     );
     $display->set('content', $root_display->get('content'));
     $display->set('hidden', $root_display->get('hidden'));
+
+    if ($root_display instanceof LayoutBuilderEntityViewDisplay) {
+      $layout_builder_settings = $root_display->getThirdPartySettings('layout_builder');
+      foreach ($layout_builder_settings as $setting_key => $setting) {
+        $display->setThirdPartySetting('layout_builder', $setting_key, $setting);
+      }
+    }
```
mglaman, line 228: This ensures Layout Builder settings are copied over.

src/Routing/RouteSubscriber.php
```diff
@@ -19,20 +20,36 @@ final class RouteSubscriber extends RouteSubscriberBase {
...
   /**
    * {@inheritdoc}
    */
   protected function alterRoutes(RouteCollection $collection) {
+    // Only run if Field UI is installed.
+    if (!$this->moduleHandler->moduleExists('field_ui')) {
+      return;
+    }
+
+    $has_layout_builder = $this->moduleHandler->moduleExists('layout_builder');
```
mglaman, line 52: For now, and the foreseeable future, Layout Builder requires Field UI. So this is safe.

mglaman:
This is ready for Layout Builder review.

KarinG:
Ok testing this now!

KarinG:
Also documenting my testing here:

I just tested this -> working as advertised. Thank you @mglaman 👍

1. No log messages
2. No more bundles on add
3. And Layout Builder is working 🎉
image

jackrabbithanna:
Testing this .. seeing the issues previously documented being resolved so far.

I do see one new PHP notice though
I'm testing with CiviCRM Event entity type. And configure the layout builder to allow configuring layout per entity. 
Then clicking the  "Layout" tab, from an edit form, for example, /civicrm-event/2/edit  to arrive at /civicrm-event/2/layout

I get notice:
Warning: Illegal offset type in ./web/modules/contrib/civicrm_entity/src/Plugin/Field/BundleFieldItemList.php on line 60

Looks like maybe the $values array is being used as an array index for $options

KarinG:
Sorry - I don't see any notices adding an Event, adding a Layout for an Individual Event and saving it -> nothing here (and yes confirmed nothing in my error log either). My local is using PHP 7.3

Is this a blocker for you?

image

jackrabbithanna:
@mglaman 
Check out this line: https://github.com/eileenmcnaughton/civicrm_entity/blob/8.x-3.x/src/Plugin/Field/BundleFieldItemList.php#L60

Should $values array is being used as an array index for $options?

@KarinG .. if you had already started layout changes for a particular entity, you may have to "Discard changes" and then try to modify the layout for the entity again to see the notice.
Yes, if this is a bug and php notices are produced, it would be a blocker. I'm guessing this could be a quick fix though.

KarinG:
I had not already started layout changes. I tried to reproduce on a fresh instance, but I can't -> Can you give me exact steps/screenshots? Drupal and PHP version? 

@puresyntax71 - can you reproduce the notice? If either of you can reproduce this -> you may need to push a commit to fix this warning. @mglaman is swamped w/ DrupalCon this week - and am I not able to reproduce it.

KarinG:
I only get into public function`setValue` when I `+Add CiviCRM Event` and when I `Save` said Event (not when hitting the `Layout` Builder tab)-> I'm intentionally using & in the Event Title - re: transliteration.

image

jackrabbithanna:
@puresyntax71 let's handle this last fix internally then. I understand Matt being busy.

Just looking at the code, you shouldn't ever be using an array as an index in another array. 
Hence the error:
Warning: Illegal offset type in /var/www/html/demo8.skvare.com/web/modules/contrib/civicrm_entity/src/Plugin/Field/BundleFieldItemList.php on line 60

I had posted steps to replicate before, but here they are:
I'm testing with CiviCRM Event entity type. Configure the event entity type to allow configuring layout per entity.
Then clicking the "Layout" tab, from an edit form, for example, at /civicrm-event/2/edit click "Layout" tab, which will take you to /civicrm-event/2/layout
The notice will enter the log on page load

jackrabbithanna:
It could be $values is sometimes something else, maybe its not an array, IDK I've been working on Commerce payment gateway all day and must devote my attention to that.

It can at least be something sometimes that causes the illegal offset error.

KarinG:
Yeah - I read your steps but can't reproduce so was asking for more details like Drupal and PHP versions. I'm on fresh D8.9 PHP 7.3 with no modules other than CiviCRM, CiviCRM Entity and Layout Builder enabled.

PS - if you check my phpstorm screenshot you can see `$values = "conference"` for my civicrm-event example.

jackrabbithanna:
I understand Karin, yet still there's a notice, will dig into it later
This PR fixes all the things we had discussed, could be this is just separate issue.

jackrabbithanna:
Merging this as original issues are fixed and makes it this new feature very close to done. 
Thank you very much Karin and Matt!

jackrabbithanna:
The error only happens the first time editing the layout .. If I leave the layout page without discarding changes, and come back, no notice. So only the first time trying the layout tab. If I discard the layout changes, and then revisit the layout tab the notice appears again.

mglaman:
@jackrabbithanna interesting, is it on PHP7.4? I'd like to investigate. So new entity, go to Layout to make an override. Notice should fire, yeah?

puresyntax71:
Hi Matt,

Yeah, this was how I was able to reproduce it:

With a CiviCRM entity having the option "Allow each civicrm address to have its layout customized." ticked.

1. Go to /civicrm-event/<id>.
2. Click on "Layout".
3. Click on "Discard changes" and confirm.
4. Click on "Layout" again.

Yeah, I was able to get around it using this diff:

```diff
-    $entity->get($civicrm_bundle_property)->setValue($options[$values]);
+    $entity->get($civicrm_bundle_property)->setValue($options[$values[0]['value']]);
```

Yeah, unsure if this affects other parts as I wasn't able to reproduce it elsewhere.

KarinG:
Thanks for the details Arnold - but 🤔 I am still not able to reproduce this on D8.9.13/PHP7.4.12/CE Entity - (master). This module really needs a testing infra. 

And if you check my phpstorm screenshot above (when I `+Add CiviCRM Event` and when I `Save` said Event) -> using `setValue($options[$values[0]['value']]` will break things ->

Yes - with 
git diff
```
-    $entity->get($civicrm_bundle_property)->setValue($options[$values]);
+    $entity->get($civicrm_bundle_property)->setValue($options[$values[0]['value']]);
```
I get a Fatal error trying to `+Add CiviCRM Event` and that makes sense.

image

Do you perhaps have any other modules/custom code in your instance that may be affecting this? I don't even hit my breakpoints when hitting the Layout tab. I only hit them when I `+Add CiviCRM Event` and when I `Save` said Event.

jackrabbithanna:
No custom code ...
I've found that sometimes $values is a string, and sometimes it is an array, depending on context.

Final solution will take some time to sort out..

Not to worry .. it's complex, and making this thing work with Layout builder and entity browser and inline entity form and all kinds of things will just take some time. 

I don't think this latest notice thing has much to do with bundles. 

We did find one other issue with some 404s for paths:

    /admin/structure/civicrm-entity/civicrm-activity/display/default/layout/discard-changes
    /admin/structure/civicrm-entity/civicrm-activity/display/default/layout/disable

But Arnold has a patch .. we'll be making a PR soon. 

So all is pretty well, we're pushing the envelope and nobody said it would be easy.

PR #291, opened 2021-04-14
Add missing routes.

puresyntax71:
There are 404 pages when going through "Discard changes" and disabling Layout builder for entity types with bundles such as CiviCRM event.

To reproduce:

1. Go to /admin/structure/civicrm-entity/civicrm-event/display.
2. Tick "Use Layout Builder" and "Save".
3. Click on "Manage layout".
4. Click on "Save layout".
5. Click on "Manage layout".
6. Click on "Discard changes" and "Confirm". (Originally goes to a 404 page.)
7. Untick "Use Layout Builder", "Save", and "Confirm". (Originally goes to a 404 page.)

### DIFF
--- a/src/Routing/RouteSubscriber.php
+++ b/src/Routing/RouteSubscriber.php
@@ -99,6 +99,14 @@ final class RouteSubscriber extends RouteSubscriberBase {
         $field_ui_routes["layout_builder.overrides.$entity_type_id.view"] = [
           'bundle' => $entity_type_id,
         ];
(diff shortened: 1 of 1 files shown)


jackrabbithanna:
@mglaman @KarinG 
Do you mind taking a look at this PR, and the other route definitions in the changed file?
Should 'bundle' always have the $entity_type_id .. or for Activities and Events, should it have a different value?

PR #294, opened 2021-04-26
fix notice generating field metadata for text_long fields, handle non…

jackrabbithanna:
…-custom field and custom field cases

Users reported a notice on module install and cache clears. 

I had gotten a PR from somebody that fixed the issue somewhat, and later I realized we need to handle a few cases, because's CiviCRM's getfields API action returns inconsistent field data. 

This PR should handle all cases gracefully.


jackrabbithanna:
Builds upon: https://github.com/eileenmcnaughton/civicrm_entity/pull/267
And patch provided here:
https://www.drupal.org/project/civicrm_entity/issues/3209235

PR #297, opened 2021-04-28
Add in operator.

puresyntax71:
Originally the `InOperator` plugin uses the following query when a custom field accepts multiple values:

```sql
...WHERE <table> LIKE "%<value>%";
```

This was a workaround since multi-value custom fields store values in the database using `\CRM_Core_DAO::VALUE_SEPARATOR` e.g. `\CRM_Core_DAO::VALUE_SEPARATOR1\CRM_Core_DAO::VALUE_SEPARATOR2\CRM_Core_DAO::VALUE_SEPARATOR`. It wasn't a good solution though.

Patch modifies the query so it's the same as how CiviCRM does it similar to this:

```mysql
...WHERE CAST(civicrm_value_contact_detai_2.roc_6 AS BINARY) RLIKE BINARY '1|3';
```

It still needs a workaround for "not in" operator though.

### DIFF
--- a/src/Plugin/views/filter/InOperator.php
+++ b/src/Plugin/views/filter/InOperator.php
@@ -111,8 +111,6 @@ class InOperator extends BaseInOperator {
     else {
       switch ($this->operator) {
         case 'in':
-          $condition = new Condition('OR');
-
           $values = array_map(function($value) {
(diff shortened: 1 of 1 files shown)


jackrabbithanna:
This issue fixed a bug for some properties, when those properties were used as filters. Merging #297

PR #300, opened 2021-05-03
Add workaround for textarea.

puresyntax71:
The textarea field in Drupal assumes that the database columns are "<field_name>__value" and "<field_name>__format".

No error is happening if you are just displaying the field since it uses a different method i.e. it doesn't do the "normal" way of querying the column. But when it comes to other queries that involves sorting, aggregation, filtering, etc, it uses the "normal" way of querying the column e.g. it does `WHERE CONCAT_WS(' ', civicrm_case.subject, ' ', civicrm_case.details__value) LIKE "%carol%"` in this case. `civicrm_case.details__value` doesn't exist in `civicrm_case` - only `civicrm_case.details`.

Workaround I added is to have CE specify a "real field" that will be used to get the value instead of how EntityViewsData determines the column. It also unsets the field for "format".

### DIFF
--- a/src/CivicrmEntityViewsData.php
+++ b/src/CivicrmEntityViewsData.php
@@ -785,4 +785,17 @@ class CivicrmEntityViewsData extends EntityViewsData {
     }
   }
 
+  /**
+   * {@inheritdoc}
+   */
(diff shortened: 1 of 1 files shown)


jackrabbithanna:
We have been using this for several client sites, after encountering issue and PR fixes issue. Merging #300

PR #301, opened 2021-05-03
Fix notice for setting bundles

### LINKED ISSUE #292: Notice -> when 'Revert to defaults' on a previously customized Layout config
Ok - I'm finally able to reproduce the warning -> it happens when:

civicrm-event -> allow and create a customized layout for one specific civicrm-event. 
then navigate to that event -> e.g. /civicrm-event/15
and then hit Layout -> and hit Revert to defaults - Confirm

`Warning: Illegal offset type in Drupal\civicrm_entity\Plugin\Field\BundleFieldItemList->setValue() (line 60 of sites/default/modules/civicrm_entity/src/Plugin/Field/BundleFieldItemList.php).`

Screenshot ->
image

puresyntax71:
This is an attempt to try to fix #292.

### DIFF
--- a/src/Plugin/Field/BundleFieldItemList.php
+++ b/src/Plugin/Field/BundleFieldItemList.php
@@ -57,7 +57,10 @@ class BundleFieldItemList extends FieldItemList {
     }, $options);
     $options = array_flip($options);
 
-    $entity->get($civicrm_bundle_property)->setValue($options[$values]);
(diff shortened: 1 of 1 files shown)


jackrabbithanna:
We feel this fixes the notice, and causes no harm. If later we think of something better, we can improve upon this. Merging #301

PR #304, opened 2021-05-06
Fix failing Kernel tests

mglaman:
Overview
This lets us run Kernel tests. The original tests mocked data since we couldn't install CiviCRM. That's fixed now. I didn't replace the mocks for the sake of just getting them to run.

### DIFF
--- a/src/CivicrmEntityViewsData.php
+++ b/src/CivicrmEntityViewsData.php
@@ -127,6 +127,8 @@ class CivicrmEntityViewsData extends EntityViewsData {
               'entity_type' => $this->entityType->id(),
               'base field' => $this->entityType->getKey('id'),
               'field_name' => $field_name,
+              'field table' => $target_base_table,
+              'field field' => $field_name,
             ];
           }
         }
--- a/tests/src/Kernel/CivicrmEntityTestBase.php
+++ b/tests/src/Kernel/CivicrmEntityTestBase.php
@@ -3,20 +3,24 @@
 namespace Drupal\Tests\civicrm_entity\Kernel;
 
 use Drupal\civicrm_entity\CiviCrmApiInterface;
-use Drupal\civicrm_entity\SupportedEntities;
+use Drupal\Core\Database\Database;
+use Drupal\Core\DependencyInjection\ContainerBuilder;
+use Drupal\Core\DependencyInjection\ServiceModifierInterface;
 use Drupal\KernelTests\KernelTestBase;
 use Prophecy\Argument;
+use Prophecy\Promise;
 
 /**
  * Test base to aid in mocking the CiviCRM API.
  */
-abstract class CivicrmEntityTestBase extends KernelTestBase {
+abstract class CivicrmEntityTestBase extends KernelTestBase implements ServiceModifierInterface {
 
   /**
    * {@inheritdoc}
    */
   protected static $modules = [
     'system',
+    'user',
     'civicrm',
     'civicrm_entity',
     'field',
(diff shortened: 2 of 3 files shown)


tests/src/Kernel/CivicrmEntityTestBase.php
```diff
@@ -73,19 +145,22 @@ protected function mockCiviCrmApi() {
...
-    $supported_entities = SupportedEntities::getInfo();
-    foreach ($supported_entities as $civicrm_entity_info) {
-      $civicrm_entity_name = $civicrm_entity_info['civicrm entity name'];
-      if (in_array($civicrm_entity_name, ['event', 'contact'])) {
-        continue;
-      }
-      $civicrm_api_mock->getFields($civicrm_entity_name)->willReturn($this->minimalGenericGetFields());
-    }
+    $civicrm_api_mock->getFields('event')->willReturn($this->minimalGenericGetFields());
+    $civicrm_api_mock->getFields('contact')->willReturn($this->minimalGenericGetFields());
```
mglaman: This broke events and contacts 🤦

src/CivicrmEntityViewsData.php
```diff
@@ -127,6 +127,8 @@ public function getViewsData() {
               'entity_type' => $this->entityType->id(),
               'base field' => $this->entityType->getKey('id'),
               'field_name' => $field_name,
+              'field table' => $target_base_table,
+              'field field' => $field_name,
```
mglaman, line 131: The tests were failing, and other places in Drupal core have these set.

tests/src/Kernel/CivicrmEntityTestBase.php
```diff
@@ -3,20 +3,24 @@
...
  */
-abstract class CivicrmEntityTestBase extends KernelTestBase {
+abstract class CivicrmEntityTestBase extends KernelTestBase implements ServiceModifierInterface {
 
   /**
    * {@inheritdoc}
    */
   protected static $modules = [
     'system',
+    'user',
```
mglaman, line 23: We cannot use EntityKernelTestBase because it caused entity type manager to run and `SupportedEntities` accesses the database too early. And the `user.permissions` service is required.

tests/src/Kernel/CivicrmEntityTestBase.php
```diff
@@ -27,33 +31,105 @@ abstract class CivicrmEntityTestBase extends KernelTestBase {
...
+      throw new \Exception("CiviCRM appears to have already been installed. Skipping full installation.");
+    }
+
+    \Civi\Setup::dispatcher()
+      ->addListener('civi.setup.installFiles', function (\Civi\Setup\Event\InstallFilesEvent $e) use ($file_private_path) {
+        $model = $e->getModel();
+        $model->settingsPath = implode(DIRECTORY_SEPARATOR, [$this->siteDirectory, 'civicrm.settings.php']);
+        $model->templateCompilePath = implode(DIRECTORY_SEPARATOR, [$file_private_path, 'civicrm', 'templates_c']);
+      }, 900);
+
```
mglaman, line 64: See https://lab.civicrm.org/dev/core/-/issues/2590. This works around the fact the path is `vfs://` and CiviCRM prepends the root path to Drupal.

mglaman (commented):
We can get parts of `CivicrmEntityTestBase` pushed upstream to the CiviCRM module as a base kernel test.

mglaman:
So. This was surprisingly easier than I thought. We can refactor these over time to stop using mocked data but actually insert an Event and Contact into CiviCRM for testing.

KarinG:
@jackrabbithanna we have lift off on the tests. Once you have had a chance to review/merge this @mglaman will start on adding functional js views testing. 

cc: @jitendrapurohit: testing CE now 🙂🎉

src/CivicrmEntityViewsData.php
jackrabbithanna: I tried testing a View listing Events, adding relationship to Location Block, then another from Location Block to Address. 
I feel like the location block is one of these auto-generated relationships created by this code here. 
Not seeing any problems.

jackrabbithanna:
Did some testing with Views I have made for testing various functionality, the changes to CivicrmEntityViewsData don't seem to be harmful. 
Merging this. Yay tests!!!

mglaman:
Woo, thanks @jackrabbithanna ! I'll start getting more functional Views testing added in, now

PR #305, opened 2021-05-11
Add initial functional JS testing

mglaman:
Overview
Adds a test for configuring the settings form:
* Verifies you can enable entity types and see links for their collection list
* Verifies you can modify filters

Adds a basic Views test
* Verifies the entities that show up on the setup wizard
* Verifies bundles render as part of the list.

### DIFF
--- a/civicrm_entity.info.yml
+++ b/civicrm_entity.info.yml
@@ -6,4 +6,5 @@ core_version_requirement: ^8.8 || ^9
 dependencies:
   - civicrm
   - drupal:datetime
+  - drupal:filter
   - drupal:options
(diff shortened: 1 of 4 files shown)


civicrm_entity.info.yml
```diff
@@ -6,4 +6,5 @@ core_version_requirement: ^8.8 || ^9
 dependencies:
   - civicrm
   - drupal:datetime
+  - drupal:filter
```
mglaman, line 9: We call functions for this module. Generally, it should always be available. but this is a safeguard

KarinG:
@mglaman 🎉 is this ready for me to review?

mglaman:
@KarinG yep! It's a start. I need to add a new test which actually had CiviCRM data that gets displayed

jackrabbithanna:
Looking good, let's merge it!

KarinG:
Ok that was fast! 👍🎉 I was going to review it - but I’ve barely started selecting my beans for my coffee. 

I’ll still add some notes today like eg adminUser vs rootUser (I still have vivid memories of our super weird test failures) and I also want to pull this in locally.

mglaman:
@KarinG yeah, I went with `adminUser` to have a user with very minimal permissions to avoid `uid:1` and know of permission limitations

PR #306, opened 2021-05-12
Add initial Views testing

mglaman:
Overview
Begins adding Views testing. I'm taking a lead from the JSON:API ResourceTestBase class. This has all of the test methods which delegate to various abstract or empty methods. Then test classes implement these to fulfill the test.


tests/src/FunctionalJavascript/Views/CivicrmEventViewsTest.php
```diff
@@ -0,0 +1,63 @@
...
+      'is_online_registration' => 1,
+      'registration_start_date' => 20080601,
+      'registration_end_date' => '2008-10-15',
+      'max_participants' => 100,
+      'event_full_text' => 'Sorry! We are already full',
+      'is_monetary' => 0,
+      'is_active' => 1,
+      'is_show_location' => 0,
+    ]);
+  }
```
mglaman, line 53: 🧐 This isn't creating an entity that is displayed.

tests/src/FunctionalJavascript/Views/CivicrmEventViewsTest.php
```diff
@@ -0,0 +1,63 @@
...
+
+  // @todo should we have a func which returns field names and a "configureField" method to adjust it further?
+  protected function addAndConfigureFields() {
```
mglaman: One idea I have is this:

* `getEntityFieldsToAdd` returns an array like `description__value` and `end_date`. Then we can specify the label and assert it is added via the label.
* Then a `configureField` which is invoked to configure the modal, which the root test can submit and close.

mglaman:
I think I found one reason it's failing: test database prefixes are being inserted:

```
SELECT civicrm_event.id AS id
FROM
test93662637civicrm_event civicrm_event
LIMIT 11 OFFSET 0
```

mglaman:
🤔 has CiviCRM Entity + Views been tested on a site using two different databases? I don't think it works. It's using the same database connection, `default`. So if the CiviCRM tables happen to be on the same database, it's working. That's why the tests are failing.

Screen Shot 2021-05-17 at 11 11 25 PM

mglaman:
Okay: \Drupal\views\Plugin\views\query\Sql::query

```
    $options = [];
    $target = 'default';
    $key = 'default';
    // Detect an external database and set the
    if (isset($this->view->base_database)) {
      $key = $this->view->base_database;
    }

    // Set the replica target if the replica option is set
    if (!empty($this->options['replica'])) {
      $target = 'replica';
    }
```

We need `base_database` to be set.

Which leads to \Drupal\views\ViewExecutable::initQuery

```
    // Create and initialize the query object.
    $views_data = Views::viewsData()->get($this->storage->get('base_table'));
    $this->storage->set('base_field', !empty($views_data['table']['base']['field']) ? $views_data['table']['base']['field'] : '');
    if (!empty($views_data['table']['base']['database'])) {
      $this->base_database = $views_data['table']['base']['database'];
    }
```

jackrabbithanna:
We use CiviCRM where its tables are in a separate database alot

We have to add prefixes in the settings.local.php like is found at : /civicrm/admin/setting/uf?reset=1
example:
```
$databases['default']['default']['prefix']= array(
  'default' => '',
  'civicrm_account_contact'                  => '`skvare8_demo_civicrm`.',
  'civicrm_account_invoice'                  => '`skvare8_demo_civicrm`.',
  'civicrm_acl'                              => '`skvare8_demo_civicrm`.',
  'civicrm_acl_cache'                        => '`skvare8_demo_civicrm`.',
  'civicrm_acl_contact_cache'                => '`skvare8_demo_civicrm`.',
  'civicrm_acl_entity_role'                  => '`skvare8_demo_civicrm`.',
  'civicrm_action_log'                       => '`skvare8_demo_civicrm`.',
```
where skvare8_demo_civicrm is the database with the civi tables in it

I saw a change record for Drupal that per-table prefixes were deprecated, but haven't been removed yet in D8/D9, but this works for us.

mglaman:
@jackrabbithanna ah, maybe it's because in the tests the CiviCRM tables do not have prefixes (I couldn't figure out how to get CiviCRM to respect prefixes.) But that's a good thing as we're missing an important definition:

```
    $data[$views_base_table]['table']['base'] = [
      'field' => $base_field,
      'title' => $this->entityType->getLabel(),
      'cache_contexts' => $this->entityType->getListCacheContexts(),
+     'database' => drupal_valid_test_ua() ? 'civicrm_test' : 'civicrm',
    ];
```

Running tests, now.

jackrabbithanna:
Oh wow .. adding the 'database' param .. that could remove the need for the prefixes in settings.local.php for everyone? 

I was worried about that problem. If this works this is great.

mglaman:
This could be why the workaround was needed. Unfortunately, running tests takes a long time as I have Xdebug running and need to wait for the View to be configured and run. I'm about to call it night and pick this up tomorrow. Pushing latest changes.

Getting this error, though

```
Drupal\Core\Database\ConnectionNotDefinedException: The specified database connection is not defined: civicrm_test in Drupal\Core\Database\Database::openConnection() (line 364 of core/lib/Drupal/Core/Database/Database.php).
Drupal\Core\Database\Database::getConnection('default', 'civicrm_test') (Line: 1300)
Drupal\views\Plugin\views\query\Sql->query() (Line: 1447)
Drupal\views\Plugin\views\query\Sql->build(Object) (Line: 1321)
```

jackrabbithanna:
So in settings.local.php would need like this?
```
$databases['civicrm_test']['default'] = array (
  'database' => 'cividatabasename',
  'username' => 'user name',
  'password' => 'password',
  'prefix' => '',
  'host' => 'localhost',
  'port' => '3306',
  'namespace' => 'Drupal\\Core\\Database\\Driver\\mysql',
  'driver' => 'mysql',
);

```

KarinG:
@mglaman -> 🎉 to see this coming together!

If having separate db-s for Drupal (drupal_slugname) and CiviCRM (civicrm_slugname) we currently do this:

We install civicrm as per usual - and this creates new civicrm tables in the drupal_slugname db (which we clean out later) but most importantly -> it installs CiviCRM and generates a civicrm.settings.php file. Then -> we will point to an existing civicrm_slugname (moved over from a D7 site e.g.) db ->

in civicrm.settings.php (edit CIVICRM_DSN to db **civicrm_slugname**):
`    define('CIVICRM_DSN', 'mysql://slugname:password@vsql/civicrm_slugname?new_link=true');
`This will make CiviCRM work in a seperate db

Then to make Views work -> 
in settings.php -> append -> everything that's under CiviCRM -> Administer -> System Settings -> CMS Database Integration on `civicrm/admin/setting/uf?reset=1`
```
$databases['default']['default']['prefix']= array(
  'default' => '',
  'civicrm_acl'                              => '`civicrm_slugname`.',
  'civicrm_acl_cache'                        => '`civicrm_slugname`.',
...
```

That paste into settings.php is a pain as you need to paste in new tables (like e.g. when you create a new custom data set), but it works.

mglaman:
Okay, so this is a known workaround. And we can get the tests to implement the same workaround, for now. And maybe find a fix later. Like the CiviCRM module adding this automatically so Drupal knows how to communicate to that database, properly.

jackrabbithanna:
Ima play with the code in https://github.com/eileenmcnaughton/civicrm_entity/pull/306#issuecomment-842813329
We need to do away with per-table prefixes anyway. Will update later if I make a PR.

mglaman:
Well, the problem isn't per table prefixes, here. It's that Drupal doesn't know how to query the CiviCRM database. I'd need to look at civicrm_views to see how it worked around it. But the current workaround can be added in the tests, so we can move forward with this at least. And it becomes documented :)

jackrabbithanna:
Ok Matt .. to do the queries .. we are adding bits in our settings.local.php as put here: https://github.com/eileenmcnaughton/civicrm_entity/pull/306#issuecomment-842817840
We've been adding bits like that .. except 'civicrm' instead of 'civicrm_test' .. for all our sites. 
In this way we can do db_select() and other similar type direct queries if we need to for some custom features.

mglaman:
I have an idea for a BC layer, as well. Otherwise this will all come crashing down for folks without the database connection details. I think we can dynamically ensure them for Drupal.

src/CivicrmEntityViewsData.php
```diff
@@ -60,6 +60,12 @@ public function getViewsData() {
       'field' => $base_field,
       'title' => $this->entityType->getLabel(),
       'cache_contexts' => $this->entityType->getListCacheContexts(),
+      // If we're in a test site, use the civicrm_test database connection.
+      'database' => drupal_valid_test_ua() ? 'civicrm_test' : 'civicrm',
+      // Specify our own views query plugin, to support ensuring the CiviCRM
+      // database connection exists in Drupal.
+      // @see \Drupal\views\Plugin\views\display\DisplayPluginBase::getPlugin
+      'query_id' => 'civicrm_views_query',
```
mglaman, line 68: This is the magic trick.

src/Form/CivicrmEntitySettings.php
```diff
@@ -173,6 +174,8 @@ public function submitForm(array &$form, FormStateInterface $form_state) {
     $this->routeBuilder->rebuild();
     $this->localActionManager->clearCachedDefinitions();
     $this->localTaskManager->clearCachedDefinitions();
+    // @todo inject views.views_data and call \Drupal\views\ViewsData::clear().
+    Cache::invalidateTags(['views_data']);
```
mglaman: I think we can delete this.

src/Plugin/views/query/CivicrmSql.php
```diff
@@ -0,0 +1,78 @@
...
+    // This should be added into the settings.php, but we provide a backwards
+    // compatibility layer here.
+    // @todo can we get this upstream in the CiviCRM module on initialize?
+    $this->civicrm->initialize();
+    $connection_name = drupal_valid_test_ua() ? 'civicrm_test' : 'civicrm';
+    if (!Database::getConnectionInfo($connection_name)) {
+      $civicrm_connection_info = Database::convertDbUrlToConnectionInfo(CIVICRM_DSN, DRUPAL_ROOT);
+      Database::addConnectionInfo($connection_name, 'default', $civicrm_connection_info);
+    }
+    parent::init($view, $display, $options);
```
mglaman, line 67: This is the magic trick! It ensures the connection is available if it hasn't been defined in `settings.php`

src/Plugin/views/query/CivicrmSql.php
```diff
@@ -0,0 +1,78 @@
...
+  }
+
+  public function query($get_count = FALSE) {
+    return parent::query($get_count);
+  }
+
+  public function build(ViewExecutable $view) {
+    parent::build($view);
+  }
```
mglaman: debugging, needs removal.

tests/src/FunctionalJavascript/CivicrmEntityViewsTestBase.php
```diff
@@ -0,0 +1,156 @@
...
+        'namespace' => $connection['namespace'],
+        'port' => $connection['port'],
+        // CiviCRM does not use prefixes.
+        'prefix' => '',
+      ],
+      'required' => TRUE,
+    ];
+    $this->writeSettings($settings);
+    */
+  }
```
mglaman, line 70: This was the original fix. But it prevents testing the backward compatibility layer. I wasn't sure how we should document this. I also wonder if we should test this via a Kernel test or something. Where this is defined and respected.

mglaman:
I can't figure out why the description won't render. I did a quick xdebug inside of `\Drupal\views\Plugin\views\field\EntityField` and it only ran for the title field. I'll pick this up later.

Maybe it's worth tidying up for a merge to fix the table prefixes workaround automatically?

KarinG:
Hi Matt - catching up! 🙂 

I got set up and am running this on my local now [D9.1.x; PHP 7.3.x; phpunit 8.5.8]

I just tried one test -> I ran CiviCRMEventViewsTest.php -> testAddWizardValues successfully  🎉 🙂

image

KarinG:
Artifacts -> already generated a lot! Maybe a bit much. 

I just checked on the settings form testing: looks 👍

image

KarinG:
@jackrabbithanna - if you're comfortable to merge this that would be great -> then we can continue on from there.

mglaman:
@KarinG I'll make some quick adjustments and clean up and mark it was ready / non-draft

src/Form/CivicrmEntitySettings.php
mglaman: Yeah, it's not needed. Entity definitions are always present, so the Views data always exists.

src/Form/CivicrmEntitySettings.php
```diff
@@ -3,6 +3,7 @@
 namespace Drupal\civicrm_entity\Form;
 
 use Drupal\civicrm_entity\SupportedEntities;
+use Drupal\Core\Cache\Cache;
```
mglaman: Unused import

tests/src/FunctionalJavascript/CivicrmEntityViewsTestBase.php
```diff
@@ -0,0 +1,158 @@
...
+use Drupal\Core\Url;
+
+abstract class CivicrmEntityViewsTestBase extends CivicrmEntityTestBase {
+
+  protected static $modules = [
+    'views',
+    'views_ui',
+  ];
+
+  protected static $civicrmEntityTypeId = NULL;
```
mglaman: This could use documentation

tests/src/FunctionalJavascript/CivicrmEntityViewsTestBase.php
```diff
@@ -0,0 +1,158 @@
...
+
+  protected static $civicrmEntityPermissions = [];
```
mglaman: This needs documentation

tests/src/FunctionalJavascript/CivicrmEntityViewsTestBase.php
```diff
@@ -0,0 +1,158 @@
...
+    else {
+      $this->assertSession()->fieldNotExists('show[type]');
+    }
+  }
+
+  /**
+   * Tests creating a basic view with the entity type.
+   *
+   * @group debug
+   */
```
mglaman: Remove `@group` annotation

tests/src/FunctionalJavascript/CivicrmEntityViewsTestBase.php
```diff
@@ -0,0 +1,158 @@
...
+  }
+
+  // @todo testCreateViewWithFilters()
+  // @todo testCreateViewWithSorts()
+  // @todo testCreateViewWithRelationships()
+
+  abstract protected function createSampleData();
+
+  abstract protected function doSetupCreateView();
+  abstract protected function assertCreateViewResults();
```
mglaman: Document these abstract methods for folks.

mglaman (commented):
nits and fixes to make it a bit nicer if merged in.

tests/src/FunctionalJavascript/CivicrmEntityViewsTestBase.php
```diff
@@ -0,0 +1,207 @@
...
+    $page->selectFieldOption('show[wizard_key]', 'standard:' . static::$civicrmEntityTypeId);
+    $this->assertSession()->assertWaitOnAjaxRequest();
+    $page->checkField('page[create]');
+    $page->fillField('page[path]', '/' . static::$civicrmEntityTypeId);
+    $page->pressButton('Save and edit');
+    $this->assertSession()->pageTextContains('The view ' . static::$civicrmEntityTypeId . ' view has been saved.');
+    $this->assertSession()->assertWaitOnAjaxRequest();
+    $this->doSetupCreateView();
+    $this->htmlOutput();
+    $page->pressButton('Save');
```
mglaman, line 135: 😱 of course the assertions failed, I never saved the View after it was configured. Tests green(er)

src/Plugin/views/query/CivicrmSql.php
jackrabbithanna: This is great. So much magic in one PR :)

src/Plugin/views/query/CivicrmSql.php
```diff
@@ -0,0 +1,70 @@
...
+   */
+  protected $civicrm;
+
+  /**
+   * Set the CiviCRM service.
+   *
+   * @param \Drupal\civicrm\Civicrm $civicrm
+   *   The CiviCRM service.
+   *
+   * @note we use this pattern to avoid constructor overrides.
```
jackrabbithanna, line 38: I learned something here :+1:

KarinG:
@jackrabbithanna - Matt moved it from WIP. If/when you're ready to merge this - we'll move into specific Views Activities testing (to start).

jackrabbithanna:
I want to do more testing with the Views query stuff, but it looks good and should do no harm. Merging and we'll fix any nits as they come up. Want to yall to be able to proceed. Thanks alot yall. Great stuff.  Merging #306

KarinG:
👍 Thank you!

jackrabbithanna:
What am I talking about "do more testing for views" .. we're getting Views tests :) the wonderfulness of this is sinking in

PR #307, opened 2021-05-13
Add support for ContactReference data type.

puresyntax71:
Overview
Adds support for custom fields that are of type "ContactReference" that was added on CiviCRM version 5.35.

Before
Custom fields of type "ContactReference" are not displaying correctly on the view modes and views. CE treats "ContactReference" fields as type integer.

After
The fields are considered as entity reference fields. In views, the fields can be added as a new relationship to reference contacts. This works for both single and multi-value fields.

Technical Details
Treats "ContactReference" fields as entity reference fields. Adds a new `ViewsJoin` plugin which is used when adding a relationship to the custom field in views.

Comments
N/A

Release notes snippet
N/A


puresyntax71:
Yeah, I've added a couple of updates here:

* Contact reference fields support for "single-line" rows - https://take.ms/p0164. Originally, the PR only contains a "join" plugin wherein you can create a relationship to the custom field for contact references. The problem here is that the fields would be separated into multiple rows.
* Updated the filter to be similar to the [Name](https://api.drupal.org/api/drupal/core%21modules%21user%21src%21Plugin%21views%21filter%21Name.php/class/Name/9.2.x) plugin - https://take.ms/WzMbE.

Yeah, leaving it on draft as it's not fully tested.

src/Entity/FieldDefinitionProvider.php
```diff
@@ -44,6 +44,17 @@ public function getBaseFieldDefinition(array $civicrm_field) {
               $field = $this->getIntegerDefinition($civicrm_field);
             }
           }
+          elseif (isset($civicrm_field['data_type']) && $civicrm_field['data_type'] === 'ContactReference') {
+            $field = BaseFieldDefinition::create('entity_reference')
+              ->setSetting('target_type', 'civicrm_contact')
```
jitendrapurohit, line 49: Hi @puresyntax71 Thanks for the fix. I agree this is still in progress but just tried to apply it on our install and confirmed that it displays the name of the contact reference field correctly 👍 .Although, there seems to be a miss since anonymous users still see the id instead of name. I wonder if we should set the target_type in the above line to something else eg `civicrm_contact_reference` (for custom fields) and add it to `src/SupportedEntities.php` as well.

    $civicrm_entity_info['civicrm_contact_reference'] = $civicrm_entity_info['civicrm_contact'];
    $civicrm_entity_info['civicrm_contact_reference']['permissions']['view'] = ['view all contacts', 'access all custom data'];

jackrabbithanna:
This has been running in production on a site for some time. Merging #307

src/Entity/FieldDefinitionProvider.php
jitendrapurohit: Created a PR to fix the above bug - https://github.com/eileenmcnaughton/civicrm_entity/pull/378 @puresyntax71 @jackrabbithanna

PR #309, opened 2021-06-12
Fix option display for entity reference pseudoconstants

jitendrapurohit:
Overview
Fix option display for entity reference pseudo constants

Before
Entity Reference such as Membership Type ID, Financial Type ID, contribution Page id do not display any options in the view filters.

Financial Types -

image

Membership Types -

image

After
Options are displayed for entity reference fields

Financial Types -

image

Membership Types -

image

### DIFF
--- a/src/CivicrmEntityViewsData.php
+++ b/src/CivicrmEntityViewsData.php
@@ -252,6 +252,16 @@ class CivicrmEntityViewsData extends EntityViewsData {
   }
 
   /**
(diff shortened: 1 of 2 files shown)


jackrabbithanna:
Great feature to add. Local testing has no issues. Will be pushing into our other environments, but if anything comes up we'll tackle it then. 
Looks great, merging. Thank you!

Issue #310, opened 2021-06-14
Proximity Filter only shows US States


KarinG:
Problem/Motivation
----------------------------------------
![image](https://user-images.githubusercontent.com/5340555/121970748-75022380-cd34-11eb-8f3e-b370e78ab5e1.png)

Steps to reproduce
----------------------------------------
Build a Leaflet view per instructions and double check Canada is set as default Country - go back to the View -> still no Provinces available in the Select element. 

![image](https://user-images.githubusercontent.com/5340555/121970919-d5916080-cd34-11eb-8cc9-3fc9af303b31.png)

jackrabbithanna:
MR resolved issue. Closing.

PR #311, opened 2021-06-15
Fix proximity default country

puresyntax71:
Overview
Uses the default country in the configuration. Relates to #310.

Before
https://user-images.githubusercontent.com/5340555/121970748-75022380-cd34-11eb-8f3e-b370e78ab5e1.png

After
image


KarinG:
That works great! Thank you @puresyntax71 -> 

image

PR #312, opened 2021-06-24
Add test with Activity and relationship to Contact

mglaman:
Overview
----------------------------------------
Adds a FunctionalJavascript test for Activities. Also adds helpers for including Relationships.

PR #315, opened 2021-07-23
Changing 'civicrm' to 'default'

KarinG:
Makes PR 306 work D8/two db-s and setting.php prefixes and relationships involving both drupal and civicrm tables. Want to see if the test fails!


KarinG:
@mglaman -> tests are ✅ & this edit makes things work for our complicated activities/calendar/relationships with user, taxonomy and contact, assignee, etc. I realize it's not the end and all but it works and for a Thursday afternoon that's 😀

mglaman:
@KarinG and this keeps it working using the `prefix` "workaround" correct? Using prefixes to map to the secondary database CivICRM exists in?

KarinG:
Yes! 🙂

mglaman:
Since I was able to get https://github.com/eileenmcnaughton/civicrm_entity/pull/312 green, we may not need this workaround

KarinG:
Roger that!

PR #319, opened 2021-08-05
Use default timezone from settings instead.

puresyntax71:
Overview

Display of time is somewhat incorrect since it always changes depending on the logged in user's timezone setting.

I think it might be due to the change for `date_default_timezone_get()` wherein the default timezone is set depending on the [logged in user](https://api.drupal.org/api/drupal/core%21modules%21system%21src%21TimeZoneResolver.php/function/TimeZoneResolver%3A%3AgetTimeZone/8.8.x).

Before

Date fields on CiviCRM entities are changing values depending on the user's set timezone.

After

Date fields on CiviCRM entities are based on the default timezone settings and are only changed on a per user's timezone setting on display.

### DIFF
--- a/src/CiviEntityStorage.php
+++ b/src/CiviEntityStorage.php
@@ -411,7 +411,7 @@ class CiviEntityStorage extends SqlContentEntityStorage {
           }
           else {
             $datetime_format = $definition->getSetting('datetime_type') === DateTimeItem::DATETIME_TYPE_DATE ? DateTimeItemInterface::DATE_STORAGE_FORMAT : DateTimeItemInterface::DATETIME_STORAGE_FORMAT;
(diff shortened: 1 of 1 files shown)


KarinG:
Can you please open an issue for this detailing the problem?

KarinG:
Ok tested!

Drupal site timezone
created Event - Edmonton (MT) -> 14:10 (in CiviCRM)

1. Views field -> override timezone -> Halifax (AT) -> 17:10 ✅
[remove override]
2. edited user profile timezone Vancouver (PT) -> 13:10 ✅
3. swap user profile timezone Toronto (ET) -> 16:10 ✅

jackrabbithanna:
with a confirmation, and running in production sites for some time, merging #319

PR #322, opened 2021-09-16
infinite loop when checking username existance

olivierh65:
checkUserNameExists function does not distinguish between an existing username and an invalid username.
This can generate an infinite loop. Stops after 10 attempts
Also corrects the addition of the extension to the username, to increment the number added at the end and not concatenate the increments.

### DIFF
--- a/src/Plugin/RulesAction/UserCreate.php
+++ b/src/Plugin/RulesAction/UserCreate.php
@@ -21,32 +21,32 @@ use Symfony\Component\DependencyInjection\ContainerInterface;
  *      "contact_id" = @ContextDefinition("integer",
  *        label = @Translation("CiviCRM contact ID"),
  *        description = @Translation("The CiviCRM contact ID."),
- *        required = TRUE,
+ *        required = TRUE
  *      ),
  *      "is_active" = @ContextDefinition("boolean",
  *        label = @Translation("Activate account"),
  *        description = @Translation("Set to TRUE to activate account. Leave empty to not activate the account. Defaults to TRUE."),
  *        assignment_restriction = "input",
  *        default_value = TRUE,
- *        required = FALSE,
+ *        required = FALSE
  *      ),
  *      "notify" = @ContextDefinition("boolean",
  *        label = @Translation("Send account notification email"),
  *        description = @Translation("Set to TRUE to send a notification email. Leave empty to not send an account notification email."),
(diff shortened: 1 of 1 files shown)


jackrabbithanna:
Thank you, will be testing this, on test sites, and after updating the github checks versions. 
@puresyntax71 let me know if you have any thoughts

src/Plugin/RulesAction/UserCreate.php
```diff
@@ -21,28 +21,28 @@
  *      "contact_id" = @ContextDefinition("integer",
  *        label = @Translation("CiviCRM contact ID"),
  *        description = @Translation("The CiviCRM contact ID."),
- *        required = TRUE
+ *        required = TRUE,
```
puresyntax71: I'm not entirely sure but doesn't Drupal detect this as invalid?

puresyntax71:
Ohh yeah that makes sense. I think this is a good change.

src/Plugin/RulesAction/UserCreate.php
jackrabbithanna: When I look in the Rules modules at other actions, I don't see commas at the end of the last item in the context definition.

src/Plugin/RulesAction/UserCreate.php
```diff
@@ -138,8 +138,12 @@ public function doExecute($contact_id, $is_active, $notify, $signin, $format) {
     if ($this->checkUserNameExists($params, $config->userSystem)) {
       $counter = 0;
       do {
-        $params['name'] = $params['name'] . '_' . $counter++;
-      } while ($this->checkUserNameExists($params, $config->userSystem));
+        # try to add an extension to username
```
jackrabbithanna, line 141: Can you update these comments to use // instead of # .. First letter of first word capitalized, and a '.' at the end? That would conform to Drupal coding standards

jackrabbithanna:
In testing in D8 and D9, these changes work.

jackrabbithanna:
Thank you!! merging #322

src/Plugin/RulesAction/UserCreate.php
KarinG, line 141: There is a "#' here 👆

jackrabbithanna: will have to fix in a separate PR

PR #326, opened 2021-09-22
Add Drupal user exist condition

olivierh65:
Add a condition to check if a linked srupal user already exist.

image

image

### DIFF
--- a/src/Plugin/Condition/DrupalUserExist.php
+++ b/src/Plugin/Condition/DrupalUserExist.php
@@ -38,17 +38,12 @@ class DrupalUserExist extends RulesConditionBase {
    */
   protected function doEvaluate($civi_contact) {
     $id = (int)$civi_contact->get('id')->getString();
-    if (is_int($id)) {
-      $result = civicrm_api3('UFMatch', 'get', [
-        'sequential' => 1,
-        'return' => ["uf_id"],
-        'contact_id' => $id,
-      ]);
-
-      return $result['is_error'] == 0 && $result['count'] == 1;
+    try {
+      $result = \Drupal::service('civicrm_entity.api')->getSingle('UFMatch', ['sequential' => 1,'return' => ["uf_id"],'contact_id' => $id]);
+      return TRUE;
     }
-    else {
-      return false;
+    catch (\CiviCRM_API3_Exception $e) {
+      return FALSE;
     }
   }
 
--- a/src/Plugin/Condition/DrupalUserExist.php
+++ b/src/Plugin/Condition/DrupalUserExist.php
@@ -1,0 +1,55 @@
+<?php
+
+namespace Drupal\civicrm_entity\Plugin\Condition;
+
+use Drupal\rules\Core\RulesConditionBase;
+use Drupal\user\UserInterface;
+use Drupal\civicrm\Civicrm;
+
+
+/**
+ * Provides a 'Drupal linked User exist' condition.
+ *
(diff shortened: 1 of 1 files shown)


src/Plugin/Condition/DrupalUserExist.php
```diff
@@ -0,0 +1,55 @@
...
+   * @param int $contac-id
+   *   The account to check.
+   *
+   * @return bool
+   *   TRUE if the contact_id is linked to a drupal account.
+   */
+  protected function doEvaluate($civi_contact) {
+    $id = (int)$civi_contact->get('id')->getString();
+    if (is_int($id)) {
+      $result = civicrm_api3('UFMatch', 'get', [
```
jackrabbithanna: Would be better to bring the civicrm_entity.api service into tis object with dependency injection, and use it instead of the API directly.

olivierh65: That's done!

src/Plugin/Condition/DrupalUserExist.php
```diff
@@ -0,0 +1,50 @@
...
+use Drupal\rules\Core\RulesConditionBase;
+use Drupal\user\UserInterface;
+use Drupal\civicrm\Civicrm;
+
+
+/**
+ * Provides a 'Drupal linked User exist' condition.
+ *
+ * @Condition(
+ *   id = "rules_drupal_user_exist",
```
puresyntax71, line 14: Yeah, somewhat minor as but is it okay to prefix it with `civicrm_entity_` instead - similar to [this](https://github.com/eileenmcnaughton/civicrm_entity/blob/8.x-3.x/src/Plugin/RulesAction/UserCreate.php#L17)?

src/Plugin/Condition/DrupalUserExist.php
```diff
@@ -0,0 +1,50 @@
...
+   * @param int $contac-id
+   *   The account to check.
+   *
+   * @return bool
+   *   TRUE if the contact_id is linked to a drupal account.
+   */
+  protected function doEvaluate($civi_contact) {
+    $id = (int)$civi_contact->get('id')->getString();
+    try {
+      $result = \Drupal::service('civicrm_entity.api')->getSingle('UFMatch', ['sequential' => 1,'return' => ["uf_id"],'contact_id' => $id]);
```
puresyntax71, line 42: Yeah, just noting perhaps using DI here but probably can just be updated later on. I think there are other sections not yet using DI as well.

It just seems to `return TRUE` directly as well without checking the `$result`. Is it necessary to check for `$result`? I'm not sure if it's necessary though as it might just throw an exception with the API call.

puresyntax71:
Yeah, apart from the minor comments. This is a neat feature for me as well.

jackrabbithanna:
Before this can be merged, the PR will need to be updated to address puresyntax71's suggestions.

jackrabbithanna:
I found a number of issues in the code, and decided best to start over. 
Testing against Drupal 9.2 and latest dev version of Rules 

Closing in favor of https://github.com/eileenmcnaughton/civicrm_entity/pull/341

If the above PR does not meet needs we can re-visit this

jackrabbithanna:
The condition I wrote won't work for your case though. You were on the right path with the entity_reference type.

If you want to continue to update this, then rename the file and class and call it like DrupalUserExistsContactEntityReference or something along those lines.

PR #330, opened 2021-09-27
Rules updates

puresyntax71:
Replace rules deprecations.

See merge request Skvare/civicrm_entity!90Overview
This is in relation to - https://www.drupal.org/project/civicrm_entity/issues/3218770. Arguments in Rules are not visible.

Before
Arguments in the Rules actions are not visible.

After
Visible arguments in rules actions.


jackrabbithanna:
This change is required to use the Rules Action in D9, and testing in D8, it still works.

PR #333, opened 2021-10-14
Must be more specific about CiviCRM versions.

KarinG:
Before
Testing using 5.41.2 -> 4 times

After
Testing using 5.35.* 5.39.* RC and dev-master

Technical Details
Need the * or it will resolve to latest stable

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -67,6 +67,11 @@ jobs:
           composer config repositories.0 path $GITHUB_WORKSPACE
           composer config repositories.1 composer https://packages.drupal.org/8
           COMPOSER_MEMORY_LIMIT=-1 composer require drupal/core-dev-pinned:${{ matrix.drupal }} --no-suggest
+      - name: Deal with https://lab.civicrm.org/dev/drupal/-/issues/164
+        # This is temporary until no longer testing anything less than 5.41. A more sophisticated way would be to make this conditional on civi version so that for 5.41+ it doesn't do this, giving a closer approximation to real installs for 5.41+, but as far as I know this is only used by compile-lib to make some bootstrap css files.
+        run: |
+          cd ~/drupal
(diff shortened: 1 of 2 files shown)


KarinG:
Ah right it needs a fix for -> 
`#6 /home/runner/drupal/vendor/scssph in /home/runner/drupal/vendor/scssphp/scssphp/src/Compiler.php on line 553`

Coming up

KarinG:
Ok so that proved that these tests were running against 5.41 (the scss issue was fixed in 5.41).

jackrabbithanna:
Yep, good PR. Silly me updating the versions. Can merge on test pass

KarinG:
Is there something in your civicrm-entity composer.json that we hit here -> 
`    - drupal/civicrm_entity 3.0.0-beta1 requires drupal/civicrm * -> no matching package found.`

PS - these are real test failures now.

jackrabbithanna:
We hadn't had problems with the composer.json before this.

jackrabbithanna:
We have had trouble with composer conflicts for Drupal 8.9 and CiviCRM 5.39 and above. We're upgrading to Drupal 9 on sites to use 5.39 and above, mainly because it needs to be done anyway, considering D8 EOL is November 2nd, 2021
But it is still good to figure out and support D8 with tests and CE, at least for several more months.

KarinG:
Composer is correct in that there is no `drupal/civicrm` package. But I don't think that's what it choked on. I had an extra trailing comma. It's building now.

jackrabbithanna:
I guess the change to the composer.json is ok. I know it needs 5.13 or above, but honestly, people should be using at least that version already.

KarinG:
And we are ✅ and now testing on all currently relevant versions.

jackrabbithanna:
Thanks Karin!!!
Merging #333

PR #336, opened 2021-11-12
Add extra filters(type, is_active)  to relationship join

jitendrapurohit:
Overview
Add extra filters (type, is_active) to relationship join.

Before
In D7, the relationship join has few extra filters added to the on clause.

image

In D9 - these extra conditions for selecting type and state is missing.

image

After
Filters are added.

image

Technical Details
Added a new handler `civicrm_entity_civicrm_relationship` for relationship join.

### DIFF
--- a/src/CivicrmEntityViewsData.php
+++ b/src/CivicrmEntityViewsData.php
@@ -513,6 +513,15 @@ class CivicrmEntityViewsData extends EntityViewsData {
 
         break;
 
+      case 'civicrm_relationship':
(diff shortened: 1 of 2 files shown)


jackrabbithanna:
I've tested this. 
Existing Views that already had a relationship to civicrm_relationships do not break.
The filters for relationship type work, either with one or multiple selected. 
The "only show active" checkbox filter works as advertised. 

Looks great!! Merging #336

jitendrapurohit:
Thanks @jackrabbithanna 👍

PR #341, opened 2021-12-02
adding Rules Condition, Drupal user exists for contact

jackrabbithanna:
Overview
Rules condition that checks if a user exists for a CiviCRM Contact

Inspired by https://github.com/eileenmcnaughton/civicrm_entity/pull/326


jackrabbithanna:
Tested with Drupal 9.2 and "Contact was viewed" and "After Contact has been saved" reaction events. 
Merging #341

PR #345, opened 2021-12-05
Update and rename DrupalUserExist.php to DrupalUserExistsContactEntit…

olivierh65:
Add a condition to check if a linked srupal user already exist
134344497-f46cec4f-39a2-4028-953f-7405edace64e

134344352-30d1f02a-32f7-4213-9d8b-72561e1a03d7

As suggested by jackrabbithanna, rename class and file (https://github.com/eileenmcnaughton/civicrm_entity/pull/326#issuecomment-985006801)


jackrabbithanna:
This reads like it will change the file name, of the Condition I added recently, and changes the class name. 
Should be adding a new file, that had your original logic with updates. 
In testing it seemed the plugin parameters and some of the logic was different between the two.

olivierh65:
>  Should be adding a new file, that had your original logic with updates.

the file DrupalUserExist.php has been deleted and 
 the file DrupalUserExistsContactEntityReference.php has been added. In this file, the name of the class has been changed to match.

However, I noticed an error when saving the rule in data selection mode: Expected a entity:civicrm_contact data type for context CiviCRM contact entity but got a entity_reference data type instead.
image

It seems to me that this comes from the Rules module, more specifically from the function checkDataTypeCompatible.
I can a correct behavior after adding in web\modules\contrib\rules\src\Context\ContextHandlerIntegrityTrait.php file 
    // Special case any and entity target types for now.
    if ($target_type == 'any' ||
      ($target_type == 'entity' && strpos($provided->getDataType(), 'entity:') !== FALSE)  **||
       ($provided->getDataType() == 'entity_reference' && strpos($target_type, 'entity:') !== FALSE)** ) {
      return;
    }

Or is it a bad declaration in the definition of the condition 
 ```
* context_definitions = {
 * "civicrm_contact" = @ContextDefinition("entity:civicrm_contact",
```

What do you think about this?

PR #339, opened 2021-12-02
Test additions

puresyntax71:
Overview
----------------------------------------
There are a few `@todo`s in the FunctionalJavascript for views. I've implemented them.

Before
----------------------------------------
Missing tests.

After
----------------------------------------
Added tests.

Technical Details
----------------------------------------
The tests that were added were for the following:

* testViewWithFilters
* testViewWithSorts
* testViewWithArguments

These tests were implemented for the following entity types:

* CiviCRM Activity
* CiviCRM Event
* CiviCRM Address (added this) 

Yeah, I've changed `SYMFONY_DEPRECATORS_HELPER` as well since it was erroring out for D9 jobs due to deprecations.

PR #347, opened 2021-12-10
Add kernel tests for InOperator plugin.

puresyntax71:
Overview
Depends on #339. This adds kernel tests for `InOperator` in views. 

Before

No tests.

After

Add tests.

Technical Details

There seem to be a bug for the following:

* There are cases where "Is not one of"s and "Is none of"s are not working using the "InOperator". Haven't fully tested which ones though.
* If a CiviCRM entity has a `NULL` value for a custom field e.g. a new custom field is created for an "Individual" but there are already existing contacts data, those without values are not included in the views result.

I added these as scenarios although commented them out for now.

Yeah, I have moved most CiviCRM related setup as well to a trait called `CivicrmEntityTrait` as well.


jackrabbithanna:
Merging #347

PR #350, opened 2021-12-15
Use tools: composer:v2

mglaman:
Overview
----------------------------------------
Bumps to composer:v2

fixes https://github.com/eileenmcnaughton/civicrm_entity/issues/349

Technical Details
----------------------------------------
CI only, Composer is much faster due to async

PR #351, opened 2021-12-15
Use PHP 7.4 in CI

mglaman:
Overview
7.3 isn't supported anymore. Bumps to 7.4, which also has some performance improvements.


mglaman:
Setup Drupal: ~1m on 7.3, ~45s on 7.4 (still Composer 1)

So 7.4 + #350 will mean a speed up in our CI.

PR #355, opened 2021-12-30
alter views query to set civi tables fully qualified name

jackrabbithanna:
Overview
See issue: https://www.drupal.org/project/civicrm_entity/issues/3198146

In Drupal 9.3, and Civi tables in a separate database, using prefixing is deprecated, and log messages appear informing of this. 
In Drupal 10, database prefixing is likely to be removed. 

Prefixing is what has been done for many years, using in a settings.local.php
$databases['default']['default']['prefix'] array to map CiviCRM tables. 
This will need to be removed, and a database connection similar to:
```
$databases['civicrm']['default'] = array (
  'database' => 'CIVI_DB',
  'username' => 'DB_USERNAME',
  'password' => 'DB_PASSWORD',
  'prefix' => '',
  'host' => 'localhost',
  'port' => '3306',
  'namespace' => 'Drupal\\Core\\Database\\Driver\\mysql',
  'driver' => 'mysql',
);
```

Before
CiviCRM Entity 3.0 release had this capability. Unforntunately any View that started with a non-Civi base table, and then related to Civi table, would have a query error. 
It is common to have a View listing users, that then has a Views relationship to the CiviCRM Contact table
Also, adding an Entity Reference field (ERF) from any Drupal entity type, such as Content, then having a View listing content, and adding the ERF relationship, would also break.

After
Views starting from Drupal entity type base tables, that relate to Civi tables, when Civi tables in a separate DB, don't have a query error

Technical Details
This PR will add code to the civicrm_entity_views_query_alter() hook implementation, to update table names, and join table names, to the fully qualified name, including database, e.g database_name.table_name

Comments
This is just a beginning prototype to start testing. 

CE needs to continue to support when the Civi tables are in same DB as Drupal.
Does this work with Multilingual?
Test fields, sorts, filters, contextual filters, and a variety of relationships
Does anything else break that is non-Views, e.g ERF?
Do we still need the Views query plugin in ./src/Plugin/views/query/CivicrmSql.php ?
Search API integration still work?

### DIFF
--- a/civicrm_entity.module
+++ b/civicrm_entity.module
@@ -29,6 +29,7 @@ use Drupal\field\FieldConfigInterface;
 use Drupal\layout_builder\Entity\LayoutBuilderEntityViewDisplay;
 use Drupal\views\ViewExecutable;
 use Drupal\views\Plugin\views\query\QueryPluginBase;
+use Drupal\Core\Database\Database;
 
 /**
  * Implements hook_theme().
@@ -490,6 +491,19 @@ function _civicrm_entity_post_invoke($op, $storage, $entity) {
 }
 
 function civicrm_entity_views_query_alter(ViewExecutable $view, QueryPluginBase $query) {
+  // Provide fully qualified table name in all Views queries
+  $civicrm_connection_name = drupal_valid_test_ua() ? 'civicrm_test' : 'civicrm';
+  $civicrm_connection = Database::getConnection('default', $civicrm_connection_name);
+  $table_queue =& $query->getTableQueue();
+  foreach ($table_queue as $alias => &$table_info) {
+    if (strpos($table_info['table'], 'civicrm_') === 0 && strpos($table_info['table'], '.') === FALSE) {
+      $table_info['table'] = $civicrm_connection->getFullQualifiedTableName($table_info['table']);
+    }
+    if (!empty($table_info['join']->table) && strpos($table_info['join']->table, 'civicrm_') === 0 && strpos($table_info['join']->table, '.') === FALSE) {
+      $table_info['join']->table = $civicrm_connection->getFullQualifiedTableName($table_info['join']->table);
+    }
+  }
+
   \Drupal::service('civicrm')->initialize();
   $multilingual = \CRM_Core_I18n::isMultilingual();
 
--- a/civicrm_entity.module
+++ b/civicrm_entity.module
@@ -491,16 +491,20 @@ function _civicrm_entity_post_invoke($op, $storage, $entity) {
 }
 
 function civicrm_entity_views_query_alter(ViewExecutable $view, QueryPluginBase $query) {
-  // Provide fully qualified table name in all Views queries
+  // Provide fully qualified table name in all Views queries.
+  // If CiviCRM tables in a separate database.
   $civicrm_connection_name = drupal_valid_test_ua() ? 'civicrm_test' : 'civicrm';
-  $civicrm_connection = Database::getConnection('default', $civicrm_connection_name);
-  $table_queue =& $query->getTableQueue();
-  foreach ($table_queue as $alias => &$table_info) {
-    if (strpos($table_info['table'], 'civicrm_') === 0 && strpos($table_info['table'], '.') === FALSE) {
-      $table_info['table'] = $civicrm_connection->getFullQualifiedTableName($table_info['table']);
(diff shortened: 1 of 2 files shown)


jackrabbithanna:
Current tests pass
TODO:
Test a wider variety of scenarios
Update our tests to test against Drupal 9.3
Add tests related to this change.

jackrabbithanna:
Our Views query plugin: 
https://github.com/eileenmcnaughton/civicrm_entity/blob/8.x-3.x/src/Plugin/views/query/CivicrmSql.php
Is not invoked when the base table is a non-Civi table. 
One consideration will be to remove this, as the code in the civicrm_entity_views_query_alter() should do the same thing. 
query plugin is set for Civi base tables here: https://github.com/eileenmcnaughton/civicrm_entity/blob/8.x-3.x/src/CivicrmEntityViewsData.php#L69

KarinG:
Note: several test PRs not merged yet. Once they are merged you’ll need to push a commit to this PR (a re-run tests will not test against latest).

jackrabbithanna:
testing locally, removing the 'query_id' array element in https://github.com/eileenmcnaughton/civicrm_entity/blob/8.x-3.x/src/CivicrmEntityViewsData.php#L69

With this PR, things continue to work. At least with all the different tests I've performed manually.

jackrabbithanna:
Ok, tests pass, when commenting out the Views base table definition, not using the CivicrmSql.php Views query plugin. 
WIll be doing manual tests, and some automated tests for this purpose. 

@mglaman if you have any thoughts on this subject, do let me know. 

For now, people do have a patch they can apply, and test, if using composer to pull in the D.O version
https://patch-diff.githubusercontent.com/raw/eileenmcnaughton/civicrm_entity/pull/355.patch

civicrm_entity.module
```diff
@@ -490,6 +491,19 @@ function _civicrm_entity_post_invoke($op, $storage, $entity) {
...
+  $civicrm_connection = Database::getConnection('default', $civicrm_connection_name);
+  $table_queue =& $query->getTableQueue();
+  foreach ($table_queue as $alias => &$table_info) {
+    if (strpos($table_info['table'], 'civicrm_') === 0 && strpos($table_info['table'], '.') === FALSE) {
+      $table_info['table'] = $civicrm_connection->getFullQualifiedTableName($table_info['table']);
+    }
+    if (!empty($table_info['join']->table) && strpos($table_info['join']->table, 'civicrm_') === 0 && strpos($table_info['join']->table, '.') === FALSE) {
+      $table_info['join']->table = $civicrm_connection->getFullQualifiedTableName($table_info['join']->table);
+    }
+  }
```
mglaman: Isn't this doing the exact same thing the `civicrm_views_query` is doing? Why have it live in an alter instead of the query plugin

jackrabbithanna: What I found was, when the base table for the View is a non-Civi table, like users or nodes, then the civicrm_views_query plugin is not used. The two cases I studied were View of users with relationship to contact and View of nodes with a Entity Reference field, referencing contacts, with a Views relationship to join to the contact data.  Having this in hook_views_query_alter() was a way I could figure to make it work. Definitely open to better ways to do this.

jackrabbithanna: Nothing wrong with the plugin when the base table is a Civi table. But if it won't work for when base table is a Drupal table, then I figured have code only in one place.

jackrabbithanna:
This PR breaks when Civi tables in the Drupal db.

This PR breaks Views when base table is civicrm_contact and there is a relationship to Users table.

civicrm_entity.module
jackrabbithanna: Turns out the plugin is needed. If commented, with the query alters .. would still get problems with a View with base table civicrm_contact with a relationship to users .. as the bridge relationship would then start to assume the users table was in the Civi database

jackrabbithanna:
I've updated the code to handle when Civi tables are in the Drupal database. 

I've also fixed the bug for the User relationship when starting with a base table of CiviCRM contacts. 

All tests pass, and local testing is going well. 
Now an opportunity for others to test, and I will be merging this in a week unless something else is discovered, and this will go out in a new release, 3.2
3.1 was released yesterday

jackrabbithanna:
Testing has been going well. Merging this now for next release.

PR #356, opened 2021-12-30
Test Drupal 9.3 instead of 9.2, bumping one Civi version to 5.44

jackrabbithanna:
Drupal 9.2 is deprecated
Test against Drupal 9.3 and Civi 5.44


jackrabbithanna:
After all the test updates, and now against drupal 9.3, all green checkboxes. Hurray!!!

PR #357, opened 2021-12-31
Views Relationship, for Relationship contact_id_b relationship type and other options

jackrabbithanna:
Follow up for https://github.com/eileenmcnaughton/civicrm_entity/pull/336

the "Contact ID B" relationship was not getting set to use the new Views relationship handler, with the options. 

After PR, both relationships have the options.


petednz:
works for me. many thanks

PR #358, opened 2022-01-03
Issue #3256734 - Make Views Bulk Operations a Required Dependency

anaprzsiczny:
…anagomes

Overview
This is a pull request based on [this issue on Drupal.org](https://www.drupal.org/project/civicrm_entity/issues/3256734), as per the request of the maintainer.

Technical Details
This PR only adds the Views Bulk Operations Drupal module as a dependency to CiviCRM Entity.

### DIFF
--- a/civicrm_entity.info.yml
+++ b/civicrm_entity.info.yml
@@ -8,3 +8,4 @@ dependencies:
   - drupal:datetime
   - drupal:filter
   - drupal:options
+  - drupal:views_bulk_operations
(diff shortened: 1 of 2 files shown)


jackrabbithanna:
Thank you for raising this issue and providing a PR.

Part of me thinks maybe we have a sub-module for any Views Bulk Operations actions. This would make it so people are not forced to install VBO, when perhaps only a small percentage of people make use of the actions. This is what we had in the Drupal 7 version. 

There is only the one action I made as an example for how to make these currently. 

Considering that option as well.

jackrabbithanna:
Closing in favor of https://github.com/eileenmcnaughton/civicrm_entity/pull/371 for now.

PR #360, opened 2022-01-06
Use organization name instead.

puresyntax71:
Overview
The filter "Current employer" doesn't work since it's looking for the current_employer which doesn't exist. API3 uses current_employer while the actual column is organization_name. Changed "real field" to use organization_name instead.

Before
Adding the filter "Current employer" produces this error:

> SQLSTATE[42S22]: Column not found: 1054 Unknown column 'civicrm_contact.current_employer' in 'where clause': SELECT "civicrm_contact"."id" AS "id" FROM "civicrm_contact" "civicrm_contact" WHERE ((civicrm_contact.contact_type = :civicrm_contact_contact_type)) AND ("civicrm_contact"."current_employer" LIKE :db_condition_placeholder_0 ESCAPE '\\') LIMIT 11 OFFSET 0; Array ( [:civicrm_contact_contact_type] => Individual [:db_condition_placeholder_0] => Default organization )

After
Error is gone - https://take.ms/VlQwl.

### DIFF
--- a/civicrm_entity.views.inc
+++ b/civicrm_entity.views.inc
@@ -135,6 +135,8 @@ function civicrm_entity_views_data_alter(&$data) {
     'real field' => 'id',
     'field' => ['id' => 'civicrm_entity_mailing_event_opened_rate'],
   ];
+
+  $data['civicrm_contact']['current_employer']['real field'] = 'organization_name';
 }
 
 /**
--- a/civicrm_entity.views.inc
+++ b/civicrm_entity.views.inc
@@ -137,6 +137,7 @@ function civicrm_entity_views_data_alter(&$data) {
   ];
 
   $data['civicrm_contact']['current_employer']['real field'] = 'organization_name';
+  $data['civicrm_contact']['employer_id']['field']['id'] = 'standard';
 }
 
 /**


seamuslee001:
@jitendrapurohit are you able to look at this?

jackrabbithanna:
This is not right, we're going to update this PR
The Current Enployer and Employer ID filters should be joing from civicrm_contact.id = civicrm_relationship.contact_id_a
then join civicrm_relationship.contact_id_b = civicrm_contact.id where relationship type = "Employee of"
then employer id = org contact id
current employer = org organization name.

puresyntax71:
Yeah, the `organization_name` column in the database is populated when an employer is assigned to an individual. `organzation_name` column is used as the "current_employer" by the CiviCRM API.

For the "employer_id", API doesn't seem to be populating it if you ask specify it as one of the "return" parameters:

```php
$result = civicrm_api3('Contact', 'get', [
  'sequential' => 1,
  'return' => ["employer_id", "current_employer"],
  'id' => 2,
]);
```

This returns:

```json
{

    "is_error": 0,
    "version": 3,
    "count": 1,
    "id": 2,
    "values": [
        {
            "contact_id": "2",
            "current_employer": "Test organization",
            "id": "2" 
        }
    ]
}
```

There isn't a current_employer_id output. I've swapped it out to use the "Standard" plugin instead of "EntityField" so it uses the database column instead for views as a workaround.

jackrabbithanna:
@seamuslee001 and @jitendrapurohit 
We could use a review on this one.

@puresyntax71 is right . .the "organization_name" column is dual purposed. For individual contacts, the organization_name is populated with the display name the Organization contact related by relationship, "Employee of" 

Seems a bit strange/hacky to me .. but this PR should fix it.
Test fails are unrelated.

jackrabbithanna:
Test fail is unrelated, merging #360

PR #361, opened 2022-01-11
Handle contact subtype special.

puresyntax71:
Overview
Use the `civicrm_entity_in_operator` so that it's similar to Drupal 7.

In Drupal 7, the query looks like this:

```
 WHERE (( (civicrm_contact.contact_sub_type LIKE '%Alpha%' ) 
```

In Drupal 9, the query looks like this:

```
WHERE ((civicrm_contact.contact_sub_type = 'Alpha')
```

This doesn't work since contact_sub_type is stored with the `\CRM_Core_DAO::VALUE_SEPARATOR`.

Before
Query is like this `WHERE ((civicrm_contact.contact_sub_type = 'Alpha')`.

After
Query is like this `AND ((CAST(civicrm_contact.contact_sub_type AS BINARY) RLIKE BINARY 'Alpha|Beta'))`

### DIFF
--- a/src/CivicrmEntityViewsData.php
+++ b/src/CivicrmEntityViewsData.php
@@ -471,6 +471,15 @@ class CivicrmEntityViewsData extends EntityViewsData {
           ],
         ];
 
+        $field_definitions = $this->entityFieldManager->getBaseFieldDefinitions('civicrm_contact');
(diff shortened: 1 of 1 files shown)


jackrabbithanna:
Test fail is unrelated, merging #361

PR #362, opened 2022-01-11
test against Civi Core 5.45

jackrabbithanna:
Run tests against 5.45 instead of 5.44


jackrabbithanna:
So .. ESR 5.45 works, but something in master in the last 5 days breaks us.

PR #363, opened 2022-01-20
test against Civi 5.45, the next ESR

jackrabbithanna:
Test against Civi 5.45, the next ESR version


jackrabbithanna:
Merging this instead https://github.com/eileenmcnaughton/civicrm_entity/pull/362

PR for test fails coming next

PR #364, opened 2022-01-20
unset SupportedEntities entity types if API Entity.get does not include it

jackrabbithanna:
Overview
CiviCRM Core in 5.47 will move CiviGrant to an extension. 

This causes test fails because CiviGrant is not installed by the test suite. 

Also manifested an error on upgrade. 
Before
Tests fail

After
Tests pass

Technical Details
What entity types have always been hardcoded. With entity types being moved to extensions, we have to check if Civi provides that entity type via API.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -2,6 +2,7 @@ name: Integration Tests
 on:
   workflow_dispatch:
   push:
+    branches: [ 8.x-3.x ]
   pull_request:
     types: [assigned, opened, synchronize, reopened]
     branches:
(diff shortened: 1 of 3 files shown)


jackrabbithanna:
Tests pass on local build, but fail here for dev-master. 
Same Drupal version, exact same code for dev-master CiviCRM, same commit as used in the tests
Its almost like the tests are not running against what is in this MR for CE

jackrabbithanna:
Tests for master branch now pass on github. 
Will be working on the test warning that triggers a fail in a separate PR
from @puresyntax71 experiment: https://github.com/puresyntax71/civicrm_entity/pull/4/files

PR #366, opened 2022-01-31
Update test scenarios.

puresyntax71:
Overview
Wrong method visibility and expected scenarios.

Before
Test fail.

After
Test pass.


jackrabbithanna:
We're back in the green :) Merging #366

PR #368, opened 2022-01-31
Add tests for activity contact record filter.

puresyntax71:
Overview
Add a simple test for "Activity Contact" filter plugin.

Before
No test.

After
Added a simple test for the custom plugin that tests assignee_id, target_id, and source_contact_id to see if the expected results are taken.


jackrabbithanna:
Loving it! Merging #368

PR #372, opened 2022-03-08
check if entity type is available to integrate

jackrabbithanna:
Overview
In Civicrm 5.47, CiviGrant is moved to extension
We need to check to make sure the APi entity is available to include in those entity types integrated with Drupal


jackrabbithanna:
Backport of this from the D9 version: https://github.com/eileenmcnaughton/civicrm_entity/pull/364

PR #374, opened 2022-04-07
Change values for default value.

puresyntax71:
Overview
The description for is_active is somewhat confusing. Made it consistent with the rest of the options - `signin` and `notify`.

* "FALSE", empty, or "0" is equivalent to `FALSE`.
* "TRUE" or any non-empty value is equivalent to `TRUE`.

Before
Conversion and description for `is_active` is confusing.

After
Made consistent with options like `signin` and `notify`.


jackrabbithanna:
Thanks! I've asked for the issue reporter to test and confirm this solves the issue for them:
https://www.drupal.org/project/civicrm_entity/issues/3273818

jackrabbithanna:
Works in our testing, merging this as an improvement.

PR #375, opened 2022-05-01
Check Drupal user by ID

olivierh65:
as a workaround for https://github.com/eileenmcnaughton/civicrm_entity/pull/345#issuecomment-1113500201
Check drupal user existance by userid (interger) rather than civicrm_contact (entity reference)


jackrabbithanna:
Going to merge this. All new functionality, won't affect anything existing.

PR #378, opened 2022-05-07
Fix loading of contact reference custom field value for anonymous users

jitendrapurohit:
Overview
Fix loading of contact reference custom field value for anonymous users.

Before
contact reference field is empty for anonymous users. To replicate -

- Create a contact reference custom field in civicrm Eg `Related Organization`.
- Create a contact with a value set for the above custom field.
- Create a drupal view using Contact entity and enable the above custom field in the list of fields. See screenshot -

image

Visit the view as admin user - The reference field is displayed correctly.

Visit the view as anonymous - The reference field is empty.

image

After
The custom field value is displayed correctly.

image

Technical Details
The reference field was fixed recently by @puresyntax71 at https://github.com/eileenmcnaughton/civicrm_entity/pull/307, which i reviewed and added a [comment](https://github.com/eileenmcnaughton/civicrm_entity/pull/307#discussion_r780664324) that it is not working for anonymous users. Fix here completes the requirement.

Comments
@puresyntax71 @jackrabbithanna thoughts?

Release notes snippet
Fixes display of contact reference custom field value for anonymous users.

### DIFF
--- a/src/Entity/FieldDefinitionProvider.php
+++ b/src/Entity/FieldDefinitionProvider.php
@@ -46,7 +46,7 @@ class FieldDefinitionProvider implements FieldDefinitionProviderInterface {
           }
           elseif (isset($civicrm_field['data_type']) && $civicrm_field['data_type'] === 'ContactReference') {
             $field = BaseFieldDefinition::create('entity_reference')
-              ->setSetting('target_type', 'civicrm_contact')
+              ->setSetting('target_type', 'civicrm_contact_reference')
               ->setSetting('handler', 'default');
 
             if (isset($civicrm_field['serialize']) && $civicrm_field['serialize']) {
--- a/src/Entity/FieldDefinitionProvider.php
+++ b/src/Entity/FieldDefinitionProvider.php
@@ -46,7 +46,7 @@ class FieldDefinitionProvider implements FieldDefinitionProviderInterface {
           }
           elseif (isset($civicrm_field['data_type']) && $civicrm_field['data_type'] === 'ContactReference') {
             $field = BaseFieldDefinition::create('entity_reference')
-              ->setSetting('target_type', 'civicrm_contact_reference')
+              ->setSetting('target_type', 'civicrm_contact')
               ->setSetting('handler', 'default');
 
             if (isset($civicrm_field['serialize']) && $civicrm_field['serialize']) {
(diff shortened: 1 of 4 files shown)


puresyntax71:
Yeah, working for me although I think this is more of a permission issue. Maybe not necessarily needed to create a new entity type. Currently, it creates duplicate entries on the "settings page":https://take.ms/r8tEvw and "views configuration":https://take.ms/CtXy9.

jackrabbithanna:
I am also not too keen on the new entity type. If there was another way, then that would be preferable. 
What about a the access() method on the EntityField class? Could we have a field plugin for these custom contact reference fields, override that method? 
https://api.drupal.org/api/drupal/core%21modules%21views%21src%21Plugin%21views%21field%21EntityField.php/function/EntityField%3A%3Aaccess/9.3.x

jitendrapurohit:
I think we already have the custom access handler for entities - https://github.com/eileenmcnaughton/civicrm_entity/blob/8.x-3.x/src/CivicrmEntityAccessHandler.php. The problem is that we dont really get any field information over here to decide if the call was made for custom reference field or a normal contact field. 

An alternate way is to use the field formatter https://api.drupal.org/api/drupal/core%21lib%21Drupal%21Core%21Field%21Plugin%21Field%21FieldFormatter%21EntityReferenceFormatterBase.php/function/EntityReferenceFormatterBase%3A%3AcheckAccess/9.0.x and override the checkAccess method instead. 

> Checks access to the given entity.
> 
> By default, entity 'view' access is checked. However, a subclass can choose to exclude certain items from entity access checking by immediately granting access.

I've just pushed another update using the ContactReferenceFormatter class for ref custom fields and it seems to work fine.

Can you please recheck this PR? @puresyntax71 @jackrabbithanna 

Note: For existing views, you need to re-add the custom field on the view.

puresyntax71:
Hi @jitendrapurohit and @jackrabbithanna 

Yeah, new changes still works for me.

Yeah, if you just enable "CiviCRM: view all contacts" permission to anonymous users, it would still work without the patch. I tested this locally as well.

jackrabbithanna:
Ok yeah, I like this approach. it also provides a model for anyone wanting to do more custom access control for these types of fields. 
Do consider this field formatter: https://github.com/eileenmcnaughton/civicrm_entity/blob/8.x-3.x/src/Plugin/Field/FieldFormatter/CustomMultiValue.php
Your field formatter may need something more to handle multi-valued custom contact reference fields. 
Thoughts about those?

jackrabbithanna:
I've seen the wisdom of this PR, merging

PR #379, opened 2022-05-11
Activity Targets not showing up.

puresyntax71:
Overview
Activity Target is showing empty in the "full" display and in views. For some reason, the name of the field is different from the actual property name. The machine name of the field is `target_contact_id` while the property name is `target_id`.

Before
Activity Target not showing up on "full" display of CiviCRM activity and in views.

After
Activity Target showing up.

Full display:

image

Views:

image

Technical Details
In CiviCRM API3, if you do this:

```php
$result = civicrm_api3('Activity', 'getfields', [
  'api_action' => "create",
]);
```

You get the following result:

```json
"target_contact_id": {
    "name": "target_id",
    "title": "Activity Target",
    "description": "Contact(s) participating in this activity.",
    "type": 1,
    "FKClassName": "CRM_Contact_DAO_Contact",
    "FKApiName": "Contact"
},
```

The difference is in the key and the property "name". Loading a CiviCRM entity uses the following:

```php
$result = civicrm_api3('Activity', 'getfields');
```

This results to:

```json
"target_contact_id": {
    "title": "Target Contact ID",
    "description": "Find activities with specified target contact.",
    "type": 1,
    "FKClassName": "CRM_Contact_DAO_Contact",
    "FKApiName": "Contact",
    "name": "target_contact_id"
},
```

The key and property "name" are the same.

The changes make the name and the key consistent instead.

Comments
Views still considers "Activity Target" as a single value field wherein the configurations for multivalue fields are not available. I haven't tested the effects of creation of CiviCRM activities although I think most use CE for displaying values instead of creating entities.

### DIFF
--- a/src/CiviCrmApi.php
+++ b/src/CiviCrmApi.php
@@ -67,7 +67,7 @@ class CiviCrmApi implements CiviCrmApiInterface {
   public function getFields($entity, $action = '') {
     $this->initialize();
     $result = civicrm_api3($entity, 'getfields', [
-      'sequential' => 1,
+      // 'sequential' => 1,
       'action' => $action,
     ]);
     return $result['values'];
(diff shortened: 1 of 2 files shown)


jackrabbithanna:
This has been working in our tests. Merging.

PR #380, opened 2022-05-18
unset API options, sort for getCount query

jackrabbithanna:
Overview
Follow up to: https://github.com/eileenmcnaughton/civicrm_entity/pull/373/files

Before
Error on pages like admin/structure/civicrm-entity/civicrm-contact

Due to 'options' => ['sort' => 'id ASC'] being added to query, for the getCount() .. which CiviCRM API throws an error

After
No error


jackrabbithanna:
Hotfix for error noticed after release. Merging.

PR #381, opened 2022-05-18
use view_mode from context, entity types with bundles, displaying rendered entity

jackrabbithanna:
Overview
When a CiviCRM entity type that has bundles (Activity, Event), and you render an entity of that type, via any view mode that is not 'default', the 'default' view mode is always what renders. 

Reported here:
https://www.drupal.org/project/civicrm_entity/issues/3247663

And by Seamus: 
https://chat.civicrm.org/civicrm/pl/8jor7k7a3bnidf5kxjwdik4dmh

Before
When a CiviCRM entity type that has bundles (Activity, Event), and you render an entity of that type, via any view mode that is not 'default', the 'default' view mode is always what renders. 
This can be rendering an entity as a view mode, in Views, or other places.

After
Entities render through the correct view mode


jackrabbithanna:
Tested this with rendering an entity with custom View mode, in Views, and for Entity reference fields renderings. Works for me. Going to merge, because I have to make a release right away anyway for a hotfix

PR #382, opened 2022-05-24
Add configuration for multisite.

puresyntax71:
Overview
This is in relation to https://www.drupal.org/project/civicrm_entity/issues/3281995.

Before
No options for Domain ID exist.

After
User can select a Domain ID to use or none at all:

image


jackrabbithanna:
This PR tests well. I think the test fail is unrelated, but we will investigate a bit. 
We'll allow a little time for any community feedback before merge.

jackrabbithanna:
This PR provides an option for the Views relationship connecting Drupal users to CiviCRM contacts. 

It allows site builders to choose which domain to use in the join connecting users to uf_match records to contacts, for sites where multiple Drupal sites are connected to the same CiviCRM instance via CiviCRM's multisite configuration. 

This has been running in a production instance for some time now.
Merging this PR.

PR #383, opened 2022-05-26
Use an "OR" operator for having multiple contact record type.

puresyntax71:
Overview
Changes the configuration for the "Activity contact type" to allow multiple values and join them using an "OR" statement. This way it can be joined to different types of contacts whether it's the "Assignee", "Source", or "Target".

Before
Configuration for Activity Contact relationship only allows a single value with a query similar to `ON civicrm_activity.id = civicrm_activity_contact.activity_id AND civicrm_activity_contact.record_type_id = '2'`.

After
Allows this configuration:

image

It produces a query similar to `ON civicrm_activity.id = civicrm_activity_contact.activity_id AND (civicrm_activity_contact.record_type_id = '2' OR civicrm_activity_contact.record_type_id = '3')`.


jackrabbithanna:
This feature has been used in production sites for some time. Merging #383

PR #384, opened 2022-05-27
Clean IDs before loading them.

puresyntax71:
Clean the IDs before loading them in `doLoadMultiple()`.

### DIFF
--- a/src/CiviEntityStorage.php
+++ b/src/CiviEntityStorage.php
@@ -171,41 +171,45 @@ class CiviEntityStorage extends SqlContentEntityStorage {
       $field_names[] = $field['name'];
     }
 
-    $options = [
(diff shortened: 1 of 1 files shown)


jackrabbithanna:
We had experienced some error when using Drupal aliases for CiviCRM Contact pages. 
To solve this, we added a call to cleadIds() a method on the base class for CiviCRM Entity content entities
https://api.drupal.org/api/drupal/core%21lib%21Drupal%21Core%21Entity%21ContentEntityStorageBase.php/function/ContentEntityStorageBase%3A%3AcleanIds/9

This PR updates CE to use this method before attempting to load entities, and reformats some code after that. 
This has been running in production for some time. Civi 5.45 and 5.51

Merging this PR.

Issue #387, opened 2022-06-24
Notice error when clearing cache


pradpnayak:
```
 [warning] Invalid argument supplied for foreach() NestedArray.php:327
 [warning] Invalid argument supplied for foreach() NestedArray.php:327
 [warning] Invalid argument supplied for foreach() NestedArray.php:327
 [warning] Invalid argument supplied for foreach() NestedArray.php:327
 [warning] Invalid argument supplied for foreach() NestedArray.php:327
 [warning] Invalid argument supplied for foreach() NestedArray.php:327
 [warning] Invalid argument supplied for foreach() NestedArray.php:327
 [warning] Invalid argument supplied for foreach() NestedArray.php:327
 [warning] Invalid argument supplied for foreach() NestedArray.php:327
 [warning] Invalid argument supplied for foreach() NestedArray.php:327
 [warning] Invalid argument supplied for foreach() NestedArray.php:327
 [warning] Invalid argument supplied for foreach() NestedArray.php:327
 [warning] Invalid argument supplied for foreach() NestedArray.php:327
 [warning] Invalid argument supplied for foreach() NestedArray.php:327
 [warning] Invalid argument supplied for foreach() NestedArray.php:327
 [warning] Invalid argument supplied for foreach() NestedArray.php:327
```

Replicate:

`drush cr`

Drupal : 9.41
CiviCRM: 5.50.3
civicrm_entity: [8.x-3.3](https://www.drupal.org/project/civicrm_entity/releases/8.x-3.3)

jackrabbithanna:
Merged, closing issue

PR #388, opened 2022-06-24
Fixed notice error

pradpnayak:
Overview
For https://github.com/eileenmcnaughton/civicrm_entity/issues/387


jackrabbithanna:
Going to merge this .. all other methods returning Views data return arrays ,this makes sense.

ghost:
Thank you for this.  I have been struggling to find the root cause of this error, and eventually found this fix.  Works fine.

PR #390, opened 2022-06-24
Add a revert patch.

puresyntax71:


jackrabbithanna:
I don't plan on merging this, we had a PR https://github.com/eileenmcnaughton/civicrm_entity/pull/381/files that was necessary for different view modes to be rendered. If problem with 381, we need to refine, not undo.

PR #377, opened 2022-05-05
Issue 3278881 - Null Valued boolean custom field

stesi561:
Overview
----------------------------------------
As per Issue 3278881 -> https://www.drupal.org/project/civicrm_entity/issues/3278881
Boolean (Yes/No) Custom fields with NULL value display falsely as NO.

Before
----------------------------------------
https://www.drupal.org/files/issues/2022-05-05/IncorrectValuesShown.png

After
----------------------------------------
https://www.drupal.org/files/issues/2022-05-05/FixedValues.png

PR #393, opened 2022-07-15
Adds break that was missed in PR377

stesi561:
Overview
Adds a break that was missing from the end of the switch context.

As per comment -> https://github.com/eileenmcnaughton/civicrm_entity/pull/377#discussion_r865907641

Before
Code execution flowed out of the switch - but if someone added another case they would need to add a break beforehand.

After
Tidy up with a break at end of switch statement.


stesi561:
Hmm might need something like this -> https://github.com/colemanw/webform_civicrm/commit/09bde1e747e90365ce70f455ce0b0c7821dc13f6

jackrabbithanna:
Test fails are unrelated. Merging #393

PR #395, opened 2022-09-20
Add workaround for contribution source not displaying.

puresyntax71:
Overview
Contribution source is not displaying correctly because of CiviCRM API inconsistency. If you do this in the CiviCRM API3 UI - https://take.ms/JwFOF, you'll see that no contribution source is being displayed.

This is what works instead:

```php
$result = civicrm_api3('Contribution', 'get', [
  'sequential' => 1,
  'return' => ["contribution_source", "id"],
  'options' => ['limit' => 2],
]);
```
I've added a workaround instead.

Before
No contribution is being displayed.

After
Contribution being displayed.


jackrabbithanna:
I do not replicate error condition in the 4.0.x branch, and I think this was fixed another way. 
Closing

PR #399, opened 2022-10-18
Improve civi case permissions.

puresyntax71:
Overview
Non admins are not able to view civi case types.

Before
Case Types are not visible to non admin roles.

After
Case type are viewable to other roles.

Technical Details
I've updated permission requirements for both Civi case and case types. For civi case, it was mostly dependent on "access all cases and activities" wherein it's required to view a case but the permissions for CE uses an "OR" condition in which case the delete operation relies on either "delete in CiviCase" or "access all cases and activities".

### DIFF
--- a/src/SupportedEntities.php
+++ b/src/SupportedEntities.php
@@ -99,13 +99,10 @@ final class SupportedEntities {
       'label property' => 'subject',
       'permissions' => [
         'view' => ['access all cases and activities'],
-        'edit' => ['access all cases and activities'],
-        'update' => ['access all cases and activities'],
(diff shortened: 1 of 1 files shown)


jackrabbithanna:
Merging this update. This changes permissions for Drupal versions of CiviCRM case create/edit/delete pages, from 'administer CiviCRM' to
'edit' => ['add cases'],
'create' => ['add cases'],
 'delete' => ['delete in CiviCase']

This would take effect for Entity Reference fields, or pages like, /civicrm-case/add,  /civicrm-case/1/edit, and /civicrm-case/1/delete

PR #400, opened 2022-11-19
Fix multi-valued custom field display

jitendrapurohit:
Overview
Custom Fields from Multiple Record Fieldsets not Displaying Correctly in Views

Before
When displaying a list of records in a view from a multiple records custom fieldset, the values for all the records for a custom field are repeated in each row, instead of giving the individual value for that field in each row. This is when the Multiple Fields Settings option "Display all values in the same row" is left unticked. If ticked, it works as expected. Please see the attached image as an example.

Strangely, this does not apply to a file type field (a custom field that allows you to upload a file) which is part of a multiple record fieldset. Nor does this field have a Multiple Field Settings option.

image

After
Fixed. Each row is displayed separately with required values.

Comments
Drupal Issue - https://www.drupal.org/project/civicrm_entity/issues/3312320

Release notes snippet
Fixes multi-valued custom field display in views.

### DIFF
--- a/src/Plugin/views/field/CustomEntityField.php
+++ b/src/Plugin/views/field/CustomEntityField.php
@@ -203,7 +203,7 @@ class CustomEntityField extends EntityField {
       ];
     }
 
-    return $items;
+    return $this->prepareItemsByDelta($items);
   }
 
   /**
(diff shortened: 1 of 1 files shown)


bmango:
Have applied the patch and can confirm this fixes the issue with no other errors.

jackrabbithanna:
Nice! Thank you, we'll test internally, but likely to merge.

jackrabbithanna:
Internal testing proves good. Merging #400

PR #401, opened 2022-11-23
Add argument validator for contact type.

puresyntax71:
Overview
Adds an argument validator for contact type.

Before
There is no way to validate argument by contact type.

After
Adds a configuration to validate by contact type:

image


jackrabbithanna:
Merging #401

PR #403, opened 2022-12-23
set the view mode to default

rafaldziubak:
Overview
Layout builder does not work for any civi entity, because the entity is rendered through `full` view_mode, even though it is not defined.

Described here:
https://www.drupal.org/project/civicrm_entity/issues/3247663#comment-14812845

Before
Entities always render through the `full` view mode.

After
If the view mode in the context is not in the list of available view_mode for civicrm_entity, it will be changed to `default`. 
E.g. if view_mode is `full` when not defined, it will be changed to `default`. **(see Technical Details)**
In summary, entities are rendered in the appropriate view mode.

Technical Details
image
On the entity page in each function across the call stack, view_mode is set to `full` and then attached to the context. 
Since 66524f6a313cf6d0e354f535e9348aa800d15a41 the view mode is set from context. While this works correctly for views(since we define a view mode in the format tab on the view configuration page), for entity pages the view_mode `full` is always set which is incorrect behavior.


jackrabbithanna:
We've had positive testing results on this change and now merging.

PR #404, opened 2023-01-11
Do not convert Drupal custom tables for CiviCRM entities in views

seamuslee001:
Overview
We have a client that has some drupal custom tables specifically around geolocation data for CiviCRM Contacts and Events. We found that they were created within the Drupal database as expected with table names like `civicrm_contact__geolocation....` but they were when used in views having the wrong database name appended this fixes that.

Before
CiviCRM database name appended to any drupal custom field tables for civicrm entities

After
CiviCRM database name not appended


jackrabbithanna:
Ok, I see that could be an issue. The solution looks appropriate. We will test and then likely merge.

jackrabbithanna:
I'm seeing no harm come from this in our set of tests for when Civi tables in separate databases. Merging #404

PR #409, opened 2023-02-17
drupal-check fixes

jackrabbithanna:
mglaman/drupal-check fixes


jackrabbithanna:
These are the ones I know I can fix. Many errors finding CiviCRM classes, and some PHPunit methods remain.

PR #413, opened 2023-03-21
Add option to disable links.

puresyntax71:
Overview
Add an option to disable links. Usually admins miss that entities have links to view, edit, or delete. This option allows admins to disable these links.

Ideally, these links can be handled using existing permissions for CiviCRM although there are certain scenarios where you'd want to enable permissions for CiviCRM. Enabling those permissions have possibilities of exposing all data to users e.g. "view all CiviCRM contacts" that can be used in webform_civicrm. Enabling that permission also enables canonical display of CiviCRM contacts.

Before
No way to hide "built-in" links.

After
Option to show or hide "built-in" links.


puresyntax71:
Closing in favor of #414.

PR #414, opened 2023-03-23
Add option to disable links.

puresyntax71:
Similar to #413.

### DIFF
--- a/civicrm_entity.install
+++ b/civicrm_entity.install
@@ -18,3 +18,24 @@ function civicrm_entity_install() {
 function civicrm_entity_update_8001() {
   module_set_weight('civicrm_entity', 1);
 }
+
+/**
+ * Enable Drupal pages for enabled entity types.
+ */
+function civicrm_entity_update_8002() {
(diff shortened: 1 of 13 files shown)


jackrabbithanna:
This feature will allow site admins to disable Drupal versions of pages / forms. 
View: /civicrm-contact/[contact-id]
Edit form: /civicrm-contact/[contact-id]/edit
Add form: /civicrm-contact/add
Delete: /civicrm-contact/delete

Admins can disable ALL of them for ALL enabled entity types with one checkbox. 
Or they can disable specific operations for specific entity types

This helps people NOT need to worry or develop custom access restrictions beyond what is default, should they choose to have them enabled. 

Default have this upgrade will be for all operations for all entity types to be enabled, to maintain existing behavior. 
Admins will have to go to the CE settings form to disable all or some of the operations for entity types, so no action necessary on upgrade to maintain default behavior. 

Requires a Drupal database update after module is updated.

jackrabbithanna:
I can't break this during testing. I think this is good new feature, and I'm going to merge and push this out in release.

PR #415, opened 2023-03-29
Add support for field_group.

puresyntax71:
Overview
Field group doesn't work with CE that have bundles.

Before
Field group doesn't work with CE that have bundles. This happens for both form and view modes.

After
Working now.


jackrabbithanna:
Merging #415

PR #416, opened 2023-04-04
Adds is Empty/NULL operators

stesi561:
Overview

Before

image

After
image

Release notes snippet


stesi561:
Credit to @jitendrapurohit  for the fix - I'm just trying to get it back upstream :)

jackrabbithanna:
How do you feel about creating a PR for the 4.0.x branch as well?

stesi561:
Yes had missed I was onwrong branch. Will be a few days sry.. bur only list.

PR #421, opened 2023-04-13
Disable creation of entity tables.

puresyntax71:
Overview
This solves https://www.drupal.org/project/civicrm_entity/issues/3299224.

Before
Tables for CiviCRM are duplicated in the Drupal database.

After
No tables created anymore.

### DIFF
--- a/src/Entity/Sql/CivicrmEntityStorageSchema.php
+++ b/src/Entity/Sql/CivicrmEntityStorageSchema.php
@@ -50,4 +50,11 @@ class CivicrmEntityStorageSchema extends SqlContentEntityStorageSchema {
     return FALSE;
   }
 
+  /**
+   * {@inheritdoc}
+   */
+  public function onFieldableEntityTypeCreate(EntityTypeInterface $entity_type, array $field_storage_definitions) {
+    return;
+  }
+
 }


adixon:
Hey @puresyntax71 - thanks for this. Reading between the lines, it looks like some part of Drupal's entity system changed somewhere, maybe in 9.5.x? I can see that you're implementing this: 

https://api.drupal.org/api/drupal/core%21lib%21Drupal%21Core%21Entity%21EntityTypeListener.php/function/EntityTypeListener%3A%3AonFieldableEntityTypeCreate/9.0.x

Do you have a reference to the Drupal changelog where this changed?

puresyntax71:
Hi @adixon 

> Do you have a reference to the Drupal changelog where this changed?

Yeah, I'm not entirely sure when this started although I tested with Drupal 9.3, 9.4, and 9.5 and I was able to reproduce it. I tested more for 9.3 and different versions of civicrm_entity ranging from the beta version.

MegaphoneJon:
Will this resolve the "entity/field definitions are out of sync" error that occurs when creating a new custom field in Civi (until you clear cache)?

jackrabbithanna:
@MegaphoneJon , no it will not. We need to react to a trigger when a custom field is added or removed, or sometimes after upgrades if schema is added/changed/removed .. and clear caches at those times.

PR #422, opened 2023-04-17
 Adds is Empty/NULL operators -> #416 against 4.0.x

stesi561:
Overview

Cherrypicked https://github.com/eileenmcnaughton/civicrm_entity/pull/416 onto 4.0.x

Before

image

After
image

Release notes snippet


stesi561:
Hmm. something went horribly wrong.

PR #423, opened 2023-04-17
Adds is Empty/NULL operators

stesi561:
Overview

Before

image

After
image

Release notes snippet

### DIFF
--- a/src/Plugin/views/filter/InOperator.php
+++ b/src/Plugin/views/filter/InOperator.php
@@ -140,21 +140,22 @@ class InOperator extends BaseInOperator {
 
   public function operators() {
     $operators = parent::operators();
-    $operators += [
-      'empty string' => [
-        'title' => $this->t('Is EMPTY/NULL'),
-        'method' => 'opEmptyString',
-        'short' => $this->t('empty string'),
-        'values' => 0,
-      ],
-      'not empty string' => [
-        'title' => $this->t('Is not EMPTY/NULL'),
-        'method' => 'opEmptyString',
-        'short' => $this->t('not empty string'),
-        'values' => 0,
-      ],
-    ];
-
+    if (!empty($this->definition['allow empty'])) {
+      $operators += [
+        'empty string' => [
+          'title' => $this->t('Is EMPTY/NULL'),
+          'method' => 'opEmptyString',
+          'short' => $this->t('empty string'),
+          'values' => 0,
+        ],
+        'not empty string' => [
+          'title' => $this->t('Is not EMPTY/NULL'),
+          'method' => 'opEmptyString',
+          'short' => $this->t('not empty string'),
+          'values' => 0,
+        ],
+      ];
+    }
     return $operators;
   }
 
--- a/src/Plugin/views/filter/InOperator.php
+++ b/src/Plugin/views/filter/InOperator.php
@@ -137,4 +137,44 @@ class InOperator extends BaseInOperator {
     }
   }
 
+  public function operators() {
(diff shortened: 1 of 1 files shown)


stesi561:
Cherry picked commit from https://github.com/eileenmcnaughton/civicrm_entity/pull/416 onto 4.0.x applied cleanly

stesi561:
Hmm. Probably need a test to go with this!

stesi561:
Hmm. -While looking at writing a test spotted this comment: https://github.com/eileenmcnaughton/civicrm_entity/blob/b97044fba0f294c8ce27e95bfc88066665736a27/tests/src/Kernel/Handler/FilterInOperatorTest.php#L235 and am now wondering if a better solution is to override the behaviour for the empty filters and have them work on NULL or '' instead of not matching against ''.

jackrabbithanna:
Finally testing this. 

Tested with "Contact SubType" filter on a test view. 
When the filter is first added, and selecting one of the new operators and clicking save
I see error:
`No valid values found on filter: CiviCRM Contact (CiviCRM Entity): Contact Subtype.`
If I select "normal" NULL, then the new operator then the error goes away

I feel like there's some small thing needed to do still to not have this error if first used by people

jackrabbithanna:
In the base class this object is inherited from, in the parent function operators
The empty / not empty operators are in a conditional like this:
`    if (!empty($this->definition['allow empty'])) {`
https://git.drupalcode.org/project/drupal/-/blob/10.1.x/core/modules/views/src/Plugin/views/filter/InOperator.php#L134

Looks like error message is generated here:
https://git.drupalcode.org/project/drupal/-/blob/10.1.x/core/modules/views/src/Plugin/views/filter/InOperator.php#L454

I see in PHP error log this error as well:
`Uncaught PHP Exception Error: "Class "Drupal\civicrm_entity\Plugin\views\filter\Condition" not found" at /var/www/web/distro10.skvare.com/web/modules/contrib/civicrm_entity/src/Plugin/views/filter/InOperator.php line 174`

I think the php file needs a "use" statement for Condition.

src/Plugin/views/filter/InOperator.php
```diff
@@ -137,4 +137,44 @@ protected function opSimple() {
...
+    if ($this->operator == 'empty string') {
+      $operator = "=";
+      $nullOP = 'IS NULL';
+    }
+    else {
+      $operator = "!=";
+      $nullOP = 'IS NOT NULL';
+    }
+
+    $condition = new Condition('OR');
```
jackrabbithanna, line 175: No "use" statement in file for this object. This throws an error to the PHP log

src/Plugin/views/filter/InOperator.php
```diff
@@ -137,4 +137,44 @@ protected function opSimple() {
     }
   }
 
+  public function operators() {
+    $operators = parent::operators();
+    $operators += [
```
jackrabbithanna: Consider wrapping this in    ` if (!empty($this->definition['allow empty'])) {`

src/Plugin/views/filter/InOperator.php
stesi561: Was confused but now it's clear! This was removed from the 4.x branch.

src/Plugin/views/filter/InOperator.php
stesi561: nice

jackrabbithanna:
Developer responded and fixed nits from feedback. Tested with contact sub-type filter in a View. Merging #423

PR #425, opened 2023-05-08
WIP: Adds Checksum validation solution for Drupal Views

stesi561:
Overview

Adds checksum access checks.

Before
No ability to grant access via checksum.

After
Adds a new Views Access Plugin that inherits from role - so we can grant access via checksum or being a logged on user.

image

Technical Details

Comments

To test create a basic view - doesn't even need to be of CiviCRM related content.
Add the access restriction CiviCRM checksum. 
A checksum can be generated via CiviCRM API 4 Explore Contact generateChecksum

Test that access is still granted via the role.
As an anon user test access is now granted via the checksum.
Add url params - cid1={contact_id}&cs={checksum}

I've followed the Webform pattern for now (with cid1) possibly this would be better with just cid?

Need some consideration of caching. I've hopefully added what is required such that this will not get cached - and as such we don't need to think about expiring when the checksum expires.

Probably needs some additional documentation

Release notes snippet

Adds a new views Access Plugin allowing you to grant access via CiviCRM checksum

### DIFF
--- a/civicrm_entity.services.yml
+++ b/civicrm_entity.services.yml
@@ -28,7 +28,7 @@ services:
 
   civicrm_entity.contact_checksum_access_checker:
     class: Drupal\civicrm_entity\Access\ContactChecksumCheckAccess
-    arguments: ['@civicrm_entity.api']
+    arguments: ['@request_stack', '@civicrm_entity.api']
     tags:
       - {name: access_check, applies_to: _civicrm_entity_checksum_access_check }
 
--- a/civicrm_entity.services.yml
+++ b/civicrm_entity.services.yml
@@ -25,3 +25,8 @@ services:
     arguments: ['@civicrm_entity.module_installer.inner', '%app.root%', '@module_handler', '@kernel', '@database', '@update.update_hook_registry']
     tags:
       - { name: service_collector, tag: 'module_install.uninstall_validator', call: addUninstallValidator }
+
+  civicrm_entity.contact_checksum_access_checker:
+    class: Drupal\civicrm_entity\Access\ContactChecksumCheckAccess
+    tags:
+      - {name: access_check, applies_to: _civicrm_entity_checksum_access_check }
--- a/civicrm_entity.services.yml
+++ b/civicrm_entity.services.yml
@@ -28,5 +28,10 @@ services:
 
   civicrm_entity.contact_checksum_access_checker:
     class: Drupal\civicrm_entity\Access\ContactChecksumCheckAccess
+    arguments: ['@civicrm_entity.api']
     tags:
       - {name: access_check, applies_to: _civicrm_entity_checksum_access_check }
+
+  civicrm_entity.views_access_contactchecksum:
+    class: Drupal\civicrm_entity\Plugin\views\Access\ContactChecksum
+    arguments: ['@request_stack', '@civicrm_entity.api']
--- a/civicrm_entity.services.yml
+++ b/civicrm_entity.services.yml
@@ -26,12 +26,13 @@ services:
     tags:
       - { name: service_collector, tag: 'module_install.uninstall_validator', call: addUninstallValidator }
 
+
   civicrm_entity.contact_checksum_access_checker:
     class: Drupal\civicrm_entity\Access\ContactChecksumCheckAccess
-    arguments: ['@request_stack', '@civicrm_entity.api']
+    autowire: true
     tags:
       - {name: access_check, applies_to: _civicrm_entity_checksum_access_check }
 
   civicrm_entity.views_access_contactchecksum:
-    class: Drupal\civicrm_entity\Plugin\views\Access\ContactChecksum
-    arguments: ['@request_stack', '@civicrm_entity.api']
+    class: Drupal\civicrm_entity\Plugin\views\access\ContactChecksum
+    autowire: true
--- a/src/Access/ContactChecksumCheckAccess.php
+++ b/src/Access/ContactChecksumCheckAccess.php
@@ -7,8 +7,10 @@ use Drupal\Core\Routing\Access\AccessInterface;
 use Drupal\Core\Session\AccountInterface;
 use Drupal\Core\Access\AccessResult;
 use Drupal\Core\Logger\LoggerChannelTrait;
+use Drupal\civicrm_entity\CiviCrmApiInterface;
 
 use Symfony\Component\HttpFoundation\RequestStack;
+
 use Symfony\Component\Routing\Route;
 
 
@@ -30,14 +32,13 @@ class ContactChecksumCheckAccess implements AccessInterface {
   /**
    * Constructs a ContactChecksumCheckAccess object.
    *
-   * @param \Symfony\Component\HttpFoundation\RequestStack $requestStack
+   * @param \Symfony\Component\HttpFoundation\RequestStack $request_stack
    *   The request stack.
    * @param \Drupal\civicrm_entity\CiviCrmApiInterface $civicrm_api
    *   The CiviCRM API bridge.
    */
-  public function __construct(RequestStack $requestStack, CiviCrmApiInterface $civicrm_api) {
-    $this->namespaces = QueryBase::getNamespaces($this);
-    $this->requestStack = $requestStack;
+  public function __construct(RequestStack $request_stack, CiviCrmApiInterface $civicrm_api) {
+    $this->requestStack = $request_stack;
     $this->civicrmApi = $civicrm_api;
   }
 
--- a/src/Access/ContactChecksumCheckAccess.php
+++ b/src/Access/ContactChecksumCheckAccess.php
@@ -6,6 +6,7 @@ use \Civi\Api4\Contact;
 use Drupal\Core\Routing\Access\AccessInterface;
 use Drupal\Core\Session\AccountInterface;
 use Drupal\Core\Access\AccessResult;
+use Drupal\Core\Logger\LoggerChannelTrait;
 
 use Symfony\Component\Routing\Route;
 
@@ -16,6 +17,7 @@ use Symfony\Component\Routing\Route;
  * Checks access for displaying views using the ContactChecksum plugin
  */
 class ContactChecksumCheckAccess implements AccessInterface {   
+  use LoggerChannelTrait;
 
   /**
    * The CiviCRM API service.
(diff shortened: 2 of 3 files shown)


src/Access/ContactChecksumCheckAccess.php
```diff
@@ -0,0 +1,57 @@
...
+
+        $cid1 = filter_var(\Drupal::request()->query->get('cid1'), FILTER_VALIDATE_INT, ['options' => ['min_range' => 1]]);
+        $checksum =  \Drupal::request()->query->get('cs');
+
+        if (empty($cid1) || empty($checksum)) {
+          \Drupal::logger('ContactChecksumCheckAccess')->info('No cid1 or cs param');
+          return AccessResult::forbidden();
+        }
+
+        $civicrmAPI = \Drupal::service('civicrm_entity.api');
```
jackrabbithanna: Can this be loaded via dependency injection? Also, could use the 'civicrm' service initialize() method

src/Plugin/views/access/ContactChecksum.php
```diff
@@ -0,0 +1,88 @@
...
+  public function access(AccountInterface $account) {
+    // Check if logged in and has access
+    $logged_in_access = parent::access($account);
+
+    if ($logged_in_access) {
+      return TRUE;
+    }
+
+    $cid1 = filter_var(\Drupal::request()->query->get('cid1'), FILTER_VALIDATE_INT, ['options' => ['min_range' => 1]]);
+    $checksum =  \Drupal::request()->query->get('cs');
```
jackrabbithanna: If possible get the request object via dependency injection

src/Access/ContactChecksumCheckAccess.php
```diff
@@ -0,0 +1,57 @@
...
+     * @return \Drupal\Core\Access\AccessResultInterface
+     *   The access result.
+     */
+    public function access(AccountInterface $account, Route $route) {
+        $options = unserialize($route->getRequirement('var_options'));
+        $account_roles = $account->getRoles();
+
+        $access_by_role = !empty(array_intersect(array_filter($options['role']), $account->getRoles()));
+        if ($access_by_role) {
+          \Drupal::logger('ContactChecksumCheckAccess')->info('Access by role');
```
jackrabbithanna: If possible, use logger by dependency injection

stesi561:
Thanks for the feedback. Agree with getting everything working using dependency injection is ideal.  Just wondering as well if this module is the correct place for this - was debating this verses a custom module or a PR on drupal civicrm module. It seems to fit in nicely with CiviCRM Entities views integration - however perhaps it makes sense that you would want to extend this to other places - i.e. have a content type that is "members only" but accessible via checksum so you can sent people smoothly to content via an email.

jackrabbithanna:
I like the feature, and would put it in CE. It provides a good feature, does it right, and CE does have the central Views functionality for CiviCRM. It also provides an example if other devs want to do a similar custom thing. There's no worries wrt upgrade of the module. Can't think of a reason not to merge it. You're choice though, if you want to turn it into a separate contrib project, that's fine too.

stesi561:
If I was to add some tests for this do you have any recommendations.

I assume we just need to do some functional JS to use browser to test that the access works?

jackrabbithanna:
Yes, we have some tests that can be used as examples to start: https://github.com/eileenmcnaughton/civicrm_entity/blob/4.0.x/tests/src/FunctionalJavascript/Views/CivicrmAddressViewsTest.php

You'd want to create one or two contacts that could be listed in a view. 
You'd want to create and login with users, one without checksum which should get access denied, and acesss one with checksum that should show results. 

I'm inclined to merge this MR soon when you get ready. I'm working on an access control blog .. for many other facets of CiviCRM Entity, and this would be great to mention too. 

I still see some places that could use dependency injection, for the request object, logging.

civicrm_entity.services.yml
```diff
@@ -25,3 +25,14 @@ services:
     arguments: ['@civicrm_entity.module_installer.inner', '%app.root%', '@module_handler', '@kernel', '@database', '@update.update_hook_registry']
     tags:
       - { name: service_collector, tag: 'module_install.uninstall_validator', call: addUninstallValidator }
+
+
+  civicrm_entity.contact_checksum_access_checker:
```
puresyntax71, line 30: The dependencies are missing here.

src/Access/ContactChecksumCheckAccess.php
```diff
@@ -0,0 +1,84 @@
...
+
+    if (empty($cid1) || empty($checksum)) {
+      $this->getlogger('ContactChecksumCheckAccess')->info('No cid1 or cs param');
+      return AccessResult::forbidden();
+    }
+
+    $this->civicrmAPI = \Drupal::service('civicrm_entity.api');
+    $this->civicrmAPI->getFields('Contact');  // This forces a call to Civicrm initialize.
+
+    $results = \Contact::validateChecksum(FALSE)
```
puresyntax71, line 78: Yeah, shouldn't this be just `Contact::validateChecksum()` i.e. without `\` at the beginning?

src/Access/ContactChecksumCheckAccess.php
```diff
@@ -0,0 +1,84 @@
...
+use Drupal\Core\Session\AccountInterface;
+use Drupal\Core\Access\AccessResult;
+use Drupal\Core\Logger\LoggerChannelTrait;
+use Drupal\civicrm_entity\CiviCrmApiInterface;
+
+use Symfony\Component\HttpFoundation\RequestStack;
+
+use Symfony\Component\Routing\Route;
+
+
```
puresyntax71, line 16: Too many spaces here.

src/Plugin/views/access/ContactChecksum.php
```diff
@@ -0,0 +1,119 @@
...
+use Drupal\Core\StringTranslation\TranslatableMarkup;
+use Drupal\user\RoleStorageInterface;
+
+use Drupal\civicrm_entity\CiviCrmApiInterface;
+
+use Symfony\Component\Routing\Route;
+
+use Drupal\user\Plugin\views\access\Role;
+use Symfony\Component\HttpFoundation\RequestStack;
+
```
puresyntax71, line 16: Too many spaces here.

src/Access/ContactChecksumCheckAccess.php
```diff
@@ -0,0 +1,84 @@
...
+   *   Run access checks for this account.
+   * @param \Symfony\Component\Routing\Route $route
+   *   The route for which an access check is being done.
+   *
+   * @return \Drupal\Core\Access\AccessResultInterface
+   *   The access result.
+   */
+  public function access(AccountInterface $account, Route $route) {
+    $options = unserialize($route->getRequirement('var_options'));
+    $account_roles = $account->getRoles();
```
puresyntax71, line 58: Unused variable.

civicrm_entity.services.yml
```diff
@@ -25,3 +25,14 @@ services:
...
+    class: Drupal\civicrm_entity\Access\ContactChecksumCheckAccess
+    autowire: true
+    tags:
+      - {name: access_check, applies_to: _civicrm_entity_checksum_access_check }
+
+  civicrm_entity.views_access_contactchecksum:
```
puresyntax71, line 36: I don't think this is needed.

src/Plugin/views/access/ContactChecksum.php
```diff
@@ -0,0 +1,119 @@
...
+  /**
+   * Constructs a ContactChecksum object.
+   *
+   * @param \Symfony\Component\HttpFoundation\RequestStack $request_stack
+   *   The request stack.
+   * @param \Drupal\civicrm_entity\CiviCrmApiInterface $civicrm_api
+   *   The CiviCRM API bridge.
+   */
+  public function __construct(RequestStack $request_stack, CiviCrmApiInterface $civicrm_api) {
```
puresyntax71, line 56: Missing the `public static function create()` implementation.

src/Plugin/views/access/ContactChecksum.php
```diff
@@ -0,0 +1,119 @@
...
+    if ($logged_in_access) {
+      return TRUE;
+    }
+    $request = $this->requestStack->getCurrentRequest();
+    $cid1 = filter_var($request->query->get('cid1'), FILTER_VALIDATE_INT, ['options' => ['min_range' => 1]]);
+    $checksum =  $request()->query->get('cs');
+
+    // Force CiviCRM to be intialized - ideally we'd use the api
+    // wrapper however it's api3 and we need api4.
+    $this->$civicrmAPI->getFields('Contact');  // This forces a call to Civicrm initialize.
```
puresyntax71, line 74: Should be `$this->civicrmApi->getFIelds('Contact');`

src/Plugin/views/access/ContactChecksum.php
```diff
@@ -0,0 +1,119 @@
...
+  public function access(AccountInterface $account) {
+    // Check if logged in and has access
+    $logged_in_access = parent::access($account);
+
+    if ($logged_in_access) {
+      return TRUE;
+    }
+    $request = $this->requestStack->getCurrentRequest();
+    $cid1 = filter_var($request->query->get('cid1'), FILTER_VALIDATE_INT, ['options' => ['min_range' => 1]]);
+    $checksum =  $request()->query->get('cs');
```
puresyntax71, line 70: Should be `$requestion->query->get('cs');`.

src/Plugin/views/access/ContactChecksum.php
```diff
@@ -0,0 +1,119 @@
...
+
+    // Force CiviCRM to be intialized - ideally we'd use the api
+    // wrapper however it's api3 and we need api4.
+    $this->$civicrmAPI->getFields('Contact');  // This forces a call to Civicrm initialize.
+
+    $results = \Contact::validateChecksum(FALSE)
```
puresyntax71, line 76: Should just be `Contact::validateChecksum(FALSE)`.

civicrm_entity.services.yml
mikeybeck: Indeed - just tested and the following was required:

>     arguments:
>       - '@request_stack'
>       - '@civicrm_entity.api'

puresyntax71:
Yeah, updated PR at #478.

jackrabbithanna:
I merged #478 so closing this PR

PR #426, opened 2023-05-09
Add support for field_group.

puresyntax71:
Overview
Similar to #415.

Before
Field group doesn't work with CE that have bundles. This happens for both form and view modes.

After
Working now.

Example edit form for event:

image

Example view form for view:

image

### DIFF
--- a/civicrm_entity.install
+++ b/civicrm_entity.install
@@ -1,0 +1,20 @@
+<?php
+
+/**
+ * @file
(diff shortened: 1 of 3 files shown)


jackrabbithanna:
This is working in my testing with both Contact and Event entity types, on forms, and displays. 

I'm going to merge this to get it into 4.0.x and the next alpha release. Thanks @puresyntax71 !!

PR #427, opened 2023-05-09
Add a check for field_group module existence.

puresyntax71:
Simply adds a check for existence of the field_group module.


jackrabbithanna:
Follow up to #426

PR #428, opened 2023-05-18
Replace isset with is_numeric in prepareItemsByDelta()

puresyntax71:
Exactly the same as #411 but for 4.0.x.


jackrabbithanna:
The version of this patch for 3.x branch has been running in production for some time. Merging #428

PR #430, opened 2023-05-30
4.0.x field mappings

puresyntax71:
Overview
Patch for https://www.drupal.org/project/civicrm_entity/issues/3203231#comment-14028098. Originally, this is specific to just the "Contact is in Trash".

There was a similar case for contribution source before although it was handled differently - #418 (more like a "field to field basis"). This is an attempt to handle all inconsistencies at once.

Before
"Contact is in Trash" not displayed properly on views. Field name is "is_deleted" although CiviCRM API uses "contact_is_deleted" to retrieve the value.

After
"Contact is in Trash" working.


jackrabbithanna:
This PR has been tested in use for some time on a couple of sites. Tests pass in a copy of it, #473

PR #431, opened 2023-06-01
make sure CiviCRM custom tables get fully qualified database name eve…

jackrabbithanna:
…n if contains double underscore

Overview
Coming from bug report: https://www.drupal.org/project/civicrm_entity/issues/3198146#comment-15086900

If a custom field table contains two underscores, the fully qualified database name is not added in Views queries, resulting in SQL error.

Before
When Civi tables in separate database from Drupal, and custom field tables contain two underscores, there's an SQL error

After
No SQL error for the conditions


jackrabbithanna:
Bug reporter says this fixes their issue. My test scenarios used previously pass. Merging #431

PR #435, opened 2023-07-22
Display yearonly custom field date without timezone modification

jitendrapurohit:
Overview
Year only custom date field is not displayed correctly in drupal view.

Before
Year only custom date field is not displayed correctly in drupal view. To replicate:

- Create a custom field of type date and format = `yy`
- Create a contact to store a year in this field, eg `2006`. The value stored in civicrm database will be `2006-01-01 00:00:00`.
- Create a drupal view and select this field to display in the result.
- Notice that the value rendered in the result is `2005` instead of `2006`.

The UTC timezone conversion converts it to `31-12-2005` date which results in the display of 2005 value instead of 2006.

After
Fixed. The yearonly date field is displayed as it is.

### DIFF
--- a/src/Plugin/views/field/CustomEntityField.php
+++ b/src/Plugin/views/field/CustomEntityField.php
@@ -295,8 +295,7 @@ class CustomEntityField extends EntityField {
     switch ($definition->getType()) {
       case 'datetime':
         if (!empty($value)) {
-          $datetime_format = $definition->getSetting('datetime_type') === DateTimeItem::DATETIME_TYPE_DATE ? DateTimeItemInterface::DATE_STORAGE_FORMAT : DateTimeItemInterface::DATETIME_STORAGE_FORMAT;
-          return (new \DateTime($value, new \DateTimeZone(date_default_timezone_get())))->setTimezone(new \DateTimeZone('UTC'))->format($datetime_format);
+          return $this->convertToUTC($definition, $value);
         }
         break;
       case 'boolean':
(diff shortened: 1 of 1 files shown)


jackrabbithanna:
@jitendrapurohit do you mind making a PR for 4.0.x branch as well? Trying for process where there first, then to 3.x. Thank you!

src/Plugin/views/field/CustomEntityField.php
```diff
@@ -310,6 +309,38 @@ protected function getItemValue($value, FieldDefinitionInterface $definition) {
...
+   * Check if date field should be converted to UTC or not.
+   *
+   * @param $definition
+   * @param $date_value
+   *
+   * @return string $datetime_value
+   */
+  public function convertToUTC($definition, $date_value) {
+    $datetime_format = $definition->getSetting('datetime_type') === DateTimeItem::DATETIME_TYPE_DATE ? DateTimeItemInterface::DATE_STORAGE_FORMAT : DateTimeItemInterface::DATETIME_STORAGE_FORMAT;
+    $default_timezone = \Drupal::config('system.date')->get('timezone.default') ?? date_default_timezone_get();
```
jackrabbithanna, line 323: Wondering why need to load the configuration for the site wide system default timezone. If users are allowed to, and they change their timezone on their user edit page, then that setting would not be respected. date_default_timezone_get() will return the proper timezone, taking into account user's timezone choice.

jackrabbithanna:
@jitendrapurohit did you have any thoughts about my comment / feedback, or my proposed slightly different PR for the 4.0.x branch? https://github.com/eileenmcnaughton/civicrm_entity/pull/439

I'm included to merge that one there, and if this one matched it would merge it too.

jitendrapurohit:
Apologies for late reply @jackrabbithanna. I've applied your alternate PR on the site where we experience this issue. Meanwhile I'll close this.

PR #436, opened 2023-07-22
Include dblocale table names in list of civicrm entity info

jitendrapurohit:
Overview
Drupal view not able to access events from participants using civicrm entity

Before
For multilingual drupal installation, the view is unable to retrieve the relationship of event table. To replicate:

- Enable multilingual support in civicrm - add FR language.
-  Create a drupal view of CiviCRM Participants.
- Try to add a relationship to CiviCRM Event table. The field is not shown since the drupal view is searching for `civicrm_event_fr_FR` or `civicrm_event_en_US` table.

After
Event field is displayed

image

### DIFF
--- a/src/SupportedEntities.php
+++ b/src/SupportedEntities.php
@@ -714,6 +714,23 @@ $civicrm_entity_info['civicrm_group_contact'] = [
       if (!in_array($entity_info['civicrm entity name'], $api_entity_types)) {
         unset($civicrm_entity_info[$entity_type]);
       }
+
+      // Insert dblocale table names
+      $multilingual = \CRM_Core_I18n::isMultilingual();
+      if ($multilingual) {
+        // @codingStandardsIgnoreStart
+        global $dbLocale;
+        // @codingStandardsIgnoreEnd
+        if ($dbLocale) {
+          $tables = \CRM_Core_I18n_Schema::schemaStructureTables();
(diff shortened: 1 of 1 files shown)


jackrabbithanna:
Oooh nice, localization improvement. Will test and review

src/SupportedEntities.php
```diff
@@ -714,6 +714,23 @@ public static function getInfo() {
       if (!in_array($entity_info['civicrm entity name'], $api_entity_types)) {
         unset($civicrm_entity_info[$entity_type]);
       }
+
+      // Insert dblocale table names
+      $multilingual = \CRM_Core_I18n::isMultilingual();
+      if ($multilingual) {
+        // @codingStandardsIgnoreStart
+        global $dbLocale;
```
jackrabbithanna, line 722: is dbLocale different if accessing via different domains or different pages? 
Seems like if so, would have to clear the Drupal caches visiting the different domains for different languages for this to work. Each time switching language would need a cache clear? Can you provide some examples of when this value is different?

jackrabbithanna:
There had been some adjustments for multi-lingual here: https://github.com/eileenmcnaughton/civicrm_entity/blob/8.x-3.x/civicrm_entity.module#L592

Perhaps the problem is there?

jackrabbithanna:
Reviewing older PRs .. merging this for a new release of 3.x

PR #439, opened 2023-08-07
fix Views field custom date field output for year only

jackrabbithanna:
Port and tweak of contributed PR https://github.com/eileenmcnaughton/civicrm_entity/pull/435

Before
Year only custom date fields for some time zones displays wrong value. 

After
Custom fields with year only format always show correctly.

### DIFF
--- a/src/Plugin/views/field/CustomEntityField.php
+++ b/src/Plugin/views/field/CustomEntityField.php
@@ -301,8 +301,7 @@ class CustomEntityField extends EntityField {
     switch ($definition->getType()) {
       case 'datetime':
         if (!empty($value)) {
-          $datetime_format = $definition->getSetting('datetime_type') === DateTimeItem::DATETIME_TYPE_DATE ? DateTimeItemInterface::DATE_STORAGE_FORMAT : DateTimeItemInterface::DATETIME_STORAGE_FORMAT;
-          return (new \DateTime($value, new \DateTimeZone(date_default_timezone_get())))->setTimezone(new \DateTimeZone('UTC'))->format($datetime_format);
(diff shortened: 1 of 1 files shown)


jackrabbithanna:
Difference between this and contributed PR for 3.x branch is to always use date_default_timezone_get() and not the system-wide timezone configuration value

src/Plugin/views/field/CustomEntityField.php
```diff
@@ -335,4 +334,40 @@ protected function getDelta($id) {
...
+   *   The field definition.
+   * @param string $date_value
+   *   The date value.
+   *
+   * @return string
+   *   The converted value.
+   */
+  public function convertToUtc(FieldDefinitionInterface $definition, $date_value) {
+    $datetime_format = $definition->getSetting('datetime_type') === DateTimeItem::DATETIME_TYPE_DATE ? DateTimeItemInterface::DATE_STORAGE_FORMAT : DateTimeItemInterface::DATETIME_STORAGE_FORMAT;
+    $default_timezone = date_default_timezone_get();
```
jackrabbithanna, line 350: This line is the difference from PR #435 for the 3.x branch

jackrabbithanna: If using the global configuration then if the per-user timezone setting is allowed, and if the user chooses a different timezone, then could have a discrepancy. Drupal manages this and gives the "right" timezone returned by  date_default_timezone_get();

jitendrapurohit:
We've applied this PR on production and can confirm this is working for year-only custom fields.

jitendrapurohit (approved):

PR #442, opened 2023-08-18
Fix column name for grouping custom fields in Views

mglaman:
Overview
Fixes the column Views uses for grouping custom fields. It was defaulting to `value`, the default column name for fields.

Technical Details
Provides a wrapped base field definition class to always return proper column names for custom fields. Modifies a View plugin method which replaces the field definition class.

Comments

Release notes snippet

### DIFF
--- a/src/Entity/CivicrmBaseFieldDefinition.php
+++ b/src/Entity/CivicrmBaseFieldDefinition.php
@@ -1,0 +1,26 @@
+<?php
+
+namespace Drupal\civicrm_entity\Entity;
+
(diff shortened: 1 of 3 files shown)


mglaman:
Okay looks like legitimate errors after I fixed my typos. Will look at it today

jackrabbithanna:
I had left this PR in one of my locals. 

I observed that with this PR, adding a Civi custom field, that is in a Civi custom field group allowing multiple values, breaks the View. 

I see errors like:
`The view test_multi_civi_fields_view failed to load an entity of type civicrm_contact at row 1 for field multi_1_20
`

PR #446, opened 2023-09-18
Ensure that only if we have a custom field that we pass it to getCust…

seamuslee001:
…omFieldMetadata

Overview
----------------------------------------
On a client site I was noticing errors in the watchdog table and the backtrace was like this and the code was essentially passing through all field_names to do a customfield.get where id = something other than a number. This aims to limit to just custom fields that we pass through to getCustomFieldMetadata

```
$backtrace = #0 /var/www/xxx/htdocs/modules/civicrm_entity/src/CiviCrmApi.php(139): CRM_Core_Error::backtrace("backtrace", TRUE)
#1 /var/www/xxx/htdocs/modules/civicrm_entity/src/Entity/CivicrmEntity.php(116): Drupal\civicrm_entity\CiviCrmApi->getCustomFieldMetadata((Array:2))
#2 /var/www/xxx/htdocs/core/lib/Drupal/Core/Entity/EntityFieldManager.php(231): Drupal\civicrm_entity\Entity\CivicrmEntity::baseFieldDefinitions(Object(Drupal\Core\Entity\ContentEntityType))
#3 /var/www/xxx/htdocs/core/lib/Drupal/Core/Entity/EntityFieldManager.php(196): Drupal\Core\Entity\EntityFieldManager->buildBaseFieldDefinitions("civicrm_action_schedule")
#4 /var/www/xxx/htdocs/modules/civicrm_entity/src/CivicrmEntityViewsData.php(104): Drupal\Core\Entity\EntityFieldManager->getBaseFieldDefinitions("civicrm_action_schedule")
#5 /var/www/xxx/htdocs/core/modules/views/views.views.inc(180): Drupal\civicrm_entity\CivicrmEntityViewsData->getViewsData()
#6 /var/www/xxx/htdocs/core/modules/views/src/ViewsData.php(236): views_views_data()
#7 /var/www/xxx/htdocs/core/lib/Drupal/Core/Extension/ModuleHandler.php(405): Drupal\views\ViewsData->Drupal\views\{closure}(Object(Closure), "views")
#8 /var/www/xxx/htdocs/core/modules/views/src/ViewsData.php(244): Drupal\Core\Extension\ModuleHandler->invokeAllWith("views_data", Object(Closure))
#9 /var/www/xxx/htdocs/core/modules/views/src/ViewsData.php(154): Drupal\views\ViewsData->getData()
[...]

PR #447, opened 2023-09-18
Ensure that only if we have a custom field that we pass it to getCust…

seamuslee001:
Overview
This is the 4.x version of https://github.com/eileenmcnaughton/civicrm_entity/pull/446

ping @jackrabbithanna


jackrabbithanna:
I'm testing this PR. Seems good for the 4.0.x branch. 4.0.x requires Drupal 10 which requires PHP8.1 so use of str_starts_with() is fine here.

jackrabbithanna:
This works in testing, merging #447

PR #448, opened 2023-09-18
fix route error on entity types with dynamic bundle type

jackrabbithanna:
Overview
On "Manage Fields" page for those entity types with dynamic bundle property, WSOD:
`TypeError: Symfony\Component\HttpFoundation\InputBag::get(): Argument #1 ($key) must be of type string, null given, called in ./web/core/lib/Drupal/Core/Routing/RouteMatch.php on line 127 in Symfony\Component\HttpFoundation\InputBag->get() (line 28 of ./vendor/symfony/http-foundation/InputBag.php).`

Before
WSOD

After
No error


jackrabbithanna:
Testing so far is good. Merging #448

PR #450, opened 2023-09-29
Add search api fix for separate civi database.

puresyntax71:
Overview
When civi database is different from Drupal database, when you build index, you get the error:

> Drupal\Core\Database\DatabaseExceptionWrapper: SQLSTATE[42S02]: Base table or view not found: 1146 Table 'drupal9.civicrm_contact' doesn't exist: SELECT "base_table"."id" AS "id" FROM "civicrm_contact" "base_table" ORDER BY "id" ASC LIMIT 100 OFFSET 0; Array ( ) in Drupal\search_api\Plugin\search_api\datasource\ContentEntity->getPartialItemIds() (line 874 of /app/web/modules/contrib/search_api/src/Plugin/search_api/datasource/ContentEntity.php).

Before
Getting errors saying table don't exist.

After
Works now even if civi database is separate from Drupal.

### DIFF
--- a/civicrm_entity.module
+++ b/civicrm_entity.module
@@ -33,6 +33,7 @@ use Drupal\Core\Database\Database;
 use Drupal\Core\Entity\Display\EntityDisplayInterface;
 use Drupal\Core\Entity\EntityInterface;
 use Drupal\Core\Form\FormStateInterface;
+use Drupal\civicrm_entity\Plugin\search_api\datasource\CivicrmEntity as DatasourceCivicrmEntity;
 
 /**
  * Implements hook_theme().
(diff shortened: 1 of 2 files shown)


kenorb:
Normally Civi credentials are in `civicrm.settings.php` and most of Civi entity classes are using `\Drupal::service('civicrm_entity.api')` to access data from the db (loading the credentials from there).
In your PR, db creds are loaded via `Database::getConnectionInfo($civicrm_connection_name)`.
Does it mean the Civi db credentials have to be defined in `$databases['civicrm']['default']` (in `settings.php`) as well?

I've tested by defining `$databases['civicrm']['default']`, seems to work fine. I can see now entities being indexed properly.

jackrabbithanna:
If your CiviCRM tables in a separate database than Drupal tables, so 2 databases, then if it required to add to settings.local.php a database connection array
`$databases['civicrm']['default'] = [`

If your CiviCRM tables are in the same database as the Drupal tables, so one database, then this should not be required. 

Views should not be working for you either, if you did not have this connection information in the settings.local.php, and 2 databases.

kenorb (approved):

jackrabbithanna:
Works in basic tests. A start to Search API integration!!! Merging #450

PR #452, opened 2023-10-11
Add state/province entity type to list of supported types

jackrabbithanna:
Overview
Add the State/Province entity type to list of supported entity types integrated into Drupal

Will be testing, and on the look out for any effects on existing Views integration, does anything change? 

Consider if "view all contacts" is the appropriate permission for the "view" operation. This is what "Country" type has now.

### DIFF
--- a/src/SupportedEntities.php
+++ b/src/SupportedEntities.php
@@ -605,6 +605,18 @@ final class SupportedEntities {
         'delete' => ['administer CiviCampaign'],
       ],
     ];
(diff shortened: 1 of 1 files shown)


jackrabbithanna:
Using this in a project, good enough so far. I may prefer a different default "view" permission. People shouldn't give user role's "View all contacts" just to view states. Most of the time in Views though, you're putting the state_province_id option group field on the View, and not a rendering of the State/Province entity itself. 

The point of this addition is to be able to reference states with entity reference fields.

PR #453, opened 2023-10-24
Add filter for state province.

puresyntax71:
Overview
There was no "State/Province" filter for in D10 version similar to D7's civicrm_handler_filter_state_multi. This provides a similar feature including the limitations. These includes:

* The "State/Province" filter also doesn't work if the country field does not exist. 
* The filter also doesn't work when there are multiple country and state province fields.

Before
No "State/Province" filter.

After
Has "State/Province" filter.

### DIFF
--- a/civicrm_entity.libraries.yml
+++ b/civicrm_entity.libraries.yml
@@ -3,3 +3,12 @@ form:
   css:
     theme:
       css/civicrm_entity.form.css: {}
+
+states:
+  version: VERSION
+  js:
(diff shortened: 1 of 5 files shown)


jackrabbithanna:
For this to work as expected, we have to disable the "Default Country" here: /civicrm/admin/setting/localization?reset=1

jackrabbithanna:
This has been running in a production environment for some time with success. Since this is a new feature that adds functionality that didn't exist, and adds value to the module, I"m going to merge it.

PR #458, opened 2023-11-29
Add reset to the file URLs generated.

puresyntax71:
Overview
When creating a view that contains file URLs, the URLs generated when visited throws this error:

> Supplied mime-type is not accepted

It looks like the URLs need `reset=1` query parameter in them.

Before
File URLs throws error.

After
No more error.


jackrabbithanna:
Working in production. Merging.

PR #459, opened 2023-12-07
loadMultiple do not check permissions in API call (WIP

jackrabbithanna:
Overview
Put check_permissions = FALSE in entity loadMultiple API calls

Before
_The current status. Please provide screenshots or gifs ([LICEcap](http://www.cockos.com/licecap/), [SilentCast](https://github.com/colinkeenan/silentcast)) where appropriate._

After
_What has been changed. Please provide screenshots or gifs ([LICEcap](http://www.cockos.com/licecap/), [SilentCast](https://github.com/colinkeenan/silentcast)) where appropriate._

Technical Details
_If the PR introduces noteworthy technical changes, please describe them here. Provide code snippets if necessary_

Comments

Release notes snippet

### DIFF
--- a/src/CiviEntityStorage.php
+++ b/src/CiviEntityStorage.php
@@ -170,6 +170,7 @@ class CiviEntityStorage extends SqlContentEntityStorage {
         'id' => ['IN' => $ids],
         'return' => $field_names,
         'options' => ['limit' => 0],
+        'check_permissions' => 0,
       ];
 
       if ($this->entityType->get('civicrm_entity') === 'participant') {


seamuslee001:
@jackrabbithanna this would essentially mean a by-passing of any ACLs that might be applied by Extensions such as the Activity Type ACL correct? For reference this is the Extension I am referring to https://civicrm.org/extensions/activity-type-acl cc @joeMurray @edsellopez

jackrabbithanna:
@seamuslee001 yes, that's the idea. In this way Views could always populate rows, and we control permission to view renderings via Drupal permissions, or custom hook_entity_access() implementations.
Drupal 8/9/10 is a little different than D7, in that Views mainly just queries for entity ids, and then loads the entities to put data in for the fields.

JoeMurray:
My sense is that when permissions are set up in CiviCRM to prevent some users from seeing certain data, it would be good for code to not default to ignoring those permissions. I'm not sure about a design pattern that requires re-implementing all of the CiviCRM permissions, created to work with Drupal, again over on the Drupal side. But perhaps I am missing something.

jackrabbithanna:
@JoeMurray That is a valid point, but CiviCRM Entity is already using its own configuration or permissions for many many things. Drupal modules are able to decide access on Views, or rendering of entities like at /civicrm-contact/1 or any of the Drupal based edit/delete forms. There's definitely use cases both ways though.
This file contains the default permissions required: https://github.com/eileenmcnaughton/civicrm_entity/blob/4.0.x/src/SupportedEntities.php#L123

ACLs are not taken into account. 

It would be possible for developers to implement hook_entity_access() and require any scheme, including CiviCRM ACLs to have access for any of the operations, for any entity type.

JoeMurray:
Thanks for your nice reply, @jackrabbithanna . 

I think this is something that is quite general, as you say. 

I'll try to initiate a broader discussion elsewhere about policies around CMS integrations and the extension of CiviCRM permissioning of access to its data to CMS permissioning of access to CiviCRM's data.

jackrabbithanna:
Yeah, and at least people being aware of how it works is good. CE has had this behavior for a long long time. The purpose is to let Drupal devs and site builders do things in a Drupal way. 
There's different expectations as well. Similar to how Seamus expected Views to just work and show results, regardless of CiviCRM permissions. We removed Views rows access control very early into the D8 version development, to mirror how it worked in Drupal 7. In D7, Views is only direct SQL queries, so no ACL or any other access control, other than the configuration of the View.

If we did change it, there would be a lot of people needing to change their sites or configurations.

jackrabbithanna:
@seamuslee001 the test failures are caused by this PR, so I'm not sure about this yet. Will loop back when I can.

jackrabbithanna:
I think this is the opposite of what you wanted Seamus. To get into access control for Views rows, see this PR that removed access control for Views rows https://github.com/eileenmcnaughton/civicrm_entity/commit/032fc7d32c78b98e436d907fff654d49c317124f

In there, would need to be able to take the entity id, and determine whether or not to include it. 

Or could get into hook_views_query_alter() implementations .. if there is an SQL way to determine if an id should be excluded.

I can reopen if desired.

PR #462, opened 2024-01-23
moving Rules integration to a submodule

jackrabbithanna:
Overview
Move Rules integration, including TypeData filters, Rules Actions, and Rules conditions to a submodule.

Technical Details
Reports of errors by users with CE installed and Rules not installed.

Release notes snippet
Change record should inform users to install CiviCRM Entity Rules module if using any of the provided conditions/actions


jackrabbithanna:
This PR was made to provide a patch for somebody that made a support request: https://www.drupal.org/project/civicrm_entity/issues/3416406
We've never experienced the reported error, and definitely have sites without the Rules module. Perhaps there is some condition for this user to experience it. 
CE could be made to require Rules, but we think it preferable for Rules to be optional. 

This change would mean people would need to install a submodule, civicrm_entity_rules IF they use any Rules conditions/actions provided by CiviCRM Entity .. they would have to read the release notes, and know to do it. 

Also considering doing something like we did for the VBO action https://github.com/eileenmcnaughton/civicrm_entity/blob/4.0.x/src/Plugin/Action/CivicrmContactAddToGroup.php#L14

PR #466, opened 2024-02-15
From #378 for 4.0.x

puresyntax71:
Similar to #378 but for 4.0.x.


jackrabbithanna:
This is an improvement. Future updates may should consider custom-ability of what permissions are required. The permissions here fit the defaults, and are safe.

PR #469, opened 2024-03-07
Ignore convert to UTC if custom field is date only

pradpnayak:
Overview
This activity in CiviCRM has custom field Date of 01/06/2023. The same is stored in db

However, when displaying within a view, this changes to 31/05/2023

Ref:https://github.com/eileenmcnaughton/civicrm_entity/pull/439/files


jackrabbithanna:
This PR under review and testing.

PR #478, opened 2024-04-27
Adds Checksum validation solution for Drupal Views

puresyntax71:
This is #425 updated.


jackrabbithanna:
This Views access control plugin allows user to configure a role(s) that are allowed to access the View
If the user does not have one of the selected role, and tries to view the View, the plugin looks for url parameters "cid1" and "cs" .. a contact id and a checksum. If the contact id and checksum validate, then the user is granted access to the View.

PR #481, opened 2024-05-10
Fix metatag not loading.

puresyntax71:
Overview
When metatag ^2 module is installed, visiting a CE page e.g. /civicrm-contact/1 throws this error:

```
Error: Call to a member function getDefaults() on null in civicrm_metatags_alter() (line 404 of modules/contrib/civicrm/civicrm.module).
```

Before
Throws an error.

After
No more error.


jackrabbithanna:
I'm merging this, so CE can be used with Metatag 2.0 in the next release

PR #483, opened 2024-05-17
adding logger parameter to ModulerInstaller service

jackrabbithanna:
Overview
per https://api.drupal.org/api/drupal/core%21lib%21Drupal%21Core%21Extension%21ModuleInstaller.php/class/ModuleInstaller/11.x
without the $logger argument is deprecated in drupal:10.1.0 and it will be required in drupal:11.0.0

Before
Error installing the module in D11

After
No error installing the module

### DIFF
--- a/civicrm_entity.info.yml
+++ b/civicrm_entity.info.yml
@@ -2,7 +2,7 @@ name: CiviCRM Entity
 type: module
 description: 'Expose CiviCRM entities as Drupal entities.'
 package: CiviCRM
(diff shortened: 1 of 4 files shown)


jackrabbithanna:
Looks like this change would break in Drupal 10.0, and requires Drupal 10.1. 

Updating .info.yml and composer.json as well.

Note in release notes that Drupal 10.1 at minimum is required.

PR #484, opened 2024-05-17
update tests for 10.3 and 11

jackrabbithanna:
Overview
Update tests to run with PHP8.3, and one test up to 10.3 and another for D11


jackrabbithanna:
I've going to merge this with the D11 tests failing. They fail due to fullcalendar_view not having a D11 version. Will update Drupal versions when 10.3 and 11 have official releases.

PR #486, opened 2024-05-21
Removing tag from contact throws an error

puresyntax71:
This is from https://www.drupal.org/project/civicrm_entity/issues/3447309. Steps to reproduce and thrown error are written from that d.o issue.


jackrabbithanna:
I don't get an error without this, but I do see that this PR is necessary, as the data is coming into the entity_id and entity_table indexes of the array, and not numeric indexes.

PR #488, opened 2024-06-11
use LoggerChannelFactoryInterface instead of default service

jackrabbithanna:
Overview
Error installing some other modules when CiviCRM Entity installed
https://www.drupal.org/project/civicrm_entity/issues/3453051

Technical Details
Accept LoggerChannelFactoryInterface instead of LoggerChannelFactory


jackrabbithanna:
Tested locally. Merging to release beta4

PR #493, opened 2024-07-24
backport PR #451 to 7.x-2.x

MegaphoneJon:
Overview
EntityTag is only non-standard with API3, API4 has fixed it.  This applies the same fix from #451 on 4.0.x to 7.x-2.x.

Before
Crash when saving in SearchKit.

After
No crash.

Release notes snippet
Fix crash when saving SearchKit on Drupal 7.


jackrabbithanna:
I'm good with this, merging. I'll put out a new release for D7

PR #494, opened 2024-07-25
custom field Views plugin, test not using createEntity

jackrabbithanna:
Overview
I saw a request for some performance optimization of Views with a large dataset
https://civicrm.stackexchange.com/questions/48376/drupal-views-using-civicrm-entity-with-relationships-added-takes-too-long-to-ren

Was provided an XHProf profile, and we discovered an inordinate amount of time spent cloning entities. 

Optimize CustomEntityField plugin.

We know that cloning the entity takes ALOT of time. 
I also wonder if we can get rid of the API call to CustomValue

Before
Molasses is faster

After
Competitive speed vs. Molasses.

### DIFF
--- a/src/CiviEntityStorage.php
+++ b/src/CiviEntityStorage.php
@@ -453,6 +453,7 @@ class CiviEntityStorage extends SqlContentEntityStorage {
           else {
             $datetime_format = $definition->getSetting('datetime_type') === DateTimeItem::DATETIME_TYPE_DATE ? DateTimeItemInterface::DATE_STORAGE_FORMAT : DateTimeItemInterface::DATETIME_STORAGE_FORMAT;
             $default_timezone = $definition->getSetting('datetime_type') === DateTimeItem::DATETIME_TYPE_DATE ? DateTimeItemInterface::STORAGE_TIMEZONE : \Drupal::config('system.date')->get('timezone.default') ?? date_default_timezone_get();
+            $item_values[$delta]['original_value'] = $item[$main_property_name];
             $datetime_value = (new \DateTime($item[$main_property_name], new \DateTimeZone($default_timezone)))->setTimezone(new \DateTimeZone('UTC'))->format($datetime_format);
             $item_values[$delta][$main_property_name] = $datetime_value;
           }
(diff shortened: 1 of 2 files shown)


jackrabbithanna:
Test passes commenting out createEntity() are not a sign all is well. 
For sure YesNo field type needs help. If add a custom field of that type to the View, then if there is no value, still "No" is rendered. 
Same for the datetime conversion done in that plugin.

There maybe other reasons why that method / entity processing needs be done.

VangelisP:
@jackrabbithanna I am facing a similar situation: My client is having around 575 customfields (!!!!) attached on the contact entity.

Whenever I ask for a simple view in Drupal 10 that shows me the name and 3-4 fields, the render time is excessively large (around 8-10 seconds) due to the fact, as you said in the stackexchange, that the entity builds with all the contact's customfields:
https://civicrm.stackexchange.com/questions/48376/drupal-views-using-civicrm-entity-with-relationships-added-takes-too-long-to-ren

I've been hunting down this for a couple of days now, trying to figure out where the bottleneck is. It does appear to be the `doLoadMultiple` that triggers a couple of apiv3 calls.

I did try your patch on `civicrm_entity` 4.0.0-beta6 and i am seriously impressed!

| Without patch 495      | With patch |
| ---------------------- | ---------- |
| Query build time: 3.13 ms    |Query build time: 2.58 ms|
| Query execute time: 1953.44 ms |Query execute time: 2074.84 ms|
| View render time: 4721.25 ms |View render time: 2536.76 ms|


This is nearly 50% render time decrease!

I was wondering, is there a way, for example, to find out what fields are being requested and fetch those back instead of rendering contact entities with all those 575 unnecessary customfields?

On a sidenote, do you see any drawbacks with your patch?

jackrabbithanna:
Hi! There were a few downsides with this patch: https://github.com/eileenmcnaughton/civicrm_entity/pull/494#issuecomment-2251149386
They are some details... 
I did try to address them. I think more testing along those lines would be good. 

Testing with 575 custom fields is exactly what this PR needs!! If it was battle hardened then would feel better about merging it.

I had worried about unintended consequences, and how to be sure, and haven't had a lot of time to devote to it. 

I don't feel like this PR is horrible though.

PR #496, opened 2024-08-09
Support for basefield overrides

puresyntax71:
Overview
This adds support for basefield overrides for entities with bundles.

Before
Base field override with bundles do not work.

After
Base field override with bundles do work. This can be tested by using https://www.drupal.org/project/base_field_override_ui to override titles and/or descriptions of base fields.


jackrabbithanna:
I've tested this and it works as advertised. Merging!

PR #497, opened 2024-08-13
Fix unknown entity type with multilingual

herbdool:
Fixes https://www.drupal.org/project/civicrm_entity/issues/3467704

### DIFF
--- a/civicrm_entity.module
+++ b/civicrm_entity.module
@@ -647,7 +647,9 @@ function civicrm_entity_views_query_alter(ViewExecutable $view, QueryPluginBase 
   $multilingual = \CRM_Core_I18n::isMultilingual();
 
   if ($multilingual) {
-    $locales = \CRM_Core_I18n::getMultilingual();
+    // @codingStandardsIgnoreStart
+    global $dbLocale;
+    // @codingStandardsIgnoreEnd
     $columns = CRM_Core_I18n_SchemaStructure::columns();
     $affectedColumns = [];
     foreach ($columns as $table => $hash) {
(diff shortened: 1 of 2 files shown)


mlutfy:
Based on code review, seems good to me (correct way to handle multi-lingual).

However, I don't have a site to test it on.

civicrm_entity.module
```diff
@@ -647,9 +647,7 @@ function civicrm_entity_views_query_alter(ViewExecutable $view, QueryPluginBase
   $multilingual = \CRM_Core_I18n::isMultilingual();
 
   if ($multilingual) {
-    // @codingStandardsIgnoreStart
-    global $dbLocale;
-    // @codingStandardsIgnoreEnd
+    $locales = \CRM_Core_I18n::getMultilingual();
```
seamuslee001: @herbdool @jackrabbithanna the changes in this file I don't think should being done

In the Supported Entities I think that makes sense as we need an entity for each locale

However in this function I think we are modifying the views query about to be executed so I would have thought we want to only use the active locale rather than all locales because i think the effect of the change in L671-L673 would mean that the last local in the list would be used rather than the active one.

E.g. if you had a list of enabled locales in CiviCRM like en_US,fr_FR,pt_PT or something and you viewed the French version of the view page (based on the url i.e. having /fr/) I think with the changes made in L671-673 and L684-686 this would mean you would only ever get the Portaguese version and also possibly you would get the completely wrong field name as well because it would seem to me that you might end up with a string like title_pt_PT_fr_FR_en_US which doesn't exist in the db as well

herbdool: That makes sense to me. I'll remove the changes in this file.

herbdool: Done

ufundo:
This change worked for me to fix issue with Multilingual CiviCRM + Webform + Drupal 10.3

herbdool:
@jackrabbithanna this PR has been tested and reviewed. I just need to rebase it and it should be good to go.

jackrabbithanna:
Merging #497

PR #500, opened 2024-09-02
update port chromedriver

KarinG:
Overview
Chromedriver uses a random port by default now.


KarinG:
Not meant to make D11 pass - but meant to solve test new failures on D10. For WFC we actually also ended up downgrading Chromedriver [similar to how Drupal core handled this issue when Olivero FunctionalJS tests started failing for D10].

jackrabbithanna:
Thanks Karin, going to merge. We're working on making the D11 tests pass and those failures are unrelated.

PR #504, opened 2024-12-05
try newer versions of drupal/civi in tests

jackrabbithanna:
Overview
Test 10.4, Civi 5.81
11.1 civi master branch
PHP 8.3 / 8.4

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -18,10 +18,10 @@ jobs:
           - drupal: '10.3.*'
             civicrm: '5.75.*'
             php: '8.1'
-          - drupal: '10.4.x-dev'
+          - drupal: '10.4.*'
             civicrm: '5.80.*'
             php: '8.3'
-          - drupal: '11.1.x-dev'
+          - drupal: '11.1.*'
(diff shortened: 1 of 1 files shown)


jackrabbithanna:
This PR to civicrm core broke tests badly: https://github.com/civicrm/civicrm-core/pull/31046
Kernel tests now fail when tried when CiviCRM Core is 5.78 or above

This file adds an event listener https://github.com/civicrm/civicrm-core/blob/master/setup/plugins/installDatabase/SyncUsers.civi-setup.php
and runs the synchonizeusers() method. 
But in the Kernel tests, the Drupal modules are not installed yet when CiviCRM is setup. 

This will require major rework of setup() in Kernel tests, and it may not be possible actually. 
We may delete them, or possibly invest in converting them to FunctionalJavascript tests.

I have no idea why synchronizeUsers() needs to be called on setup, as there are no contacts yet on install. Boooo

jackrabbithanna:
Closing in favor of https://github.com/eileenmcnaughton/civicrm_entity/pull/510 after several other test updates in other PRs

PR #505, opened 2024-12-09
Convert plugins from annotations to attributes.

puresyntax71:
Convert plugins from annotations to attributes.


jackrabbithanna:
I'm merging this. Next version of CiviCRM Entity will require Drupal 10.3

PR #507, opened 2025-01-02
Pre and Post hooks don't check if the entities are enabled for sync

sadashivdalvi:
This is a fix for https://www.drupal.org/project/civicrm_entity/issues/3496954

Here I am checking whether the component is enabled and skip if not

### DIFF
--- a/civicrm_entity.module
+++ b/civicrm_entity.module
@@ -474,7 +474,7 @@ function civicrm_entity_civicrm_pre($op, $objectName, $id, &$params) {
   }
 
   $enabled_components = \Drupal::config('civicrm_entity.settings')->get('enabled_entity_types');
-  if (!in_array($entityType, $enabled_components)) {
+  if (empty($enabled_components) || !in_array($entityType, $enabled_components)) {
     return;
   }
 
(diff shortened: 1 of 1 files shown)


jackrabbithanna:
I see a test fail, can you add a check if !empty($enabled_components) for the in_array() check?

jackrabbithanna:
Just a heads up. I'm consider a more major update to CiviCRM Entity. 
I don't like turning off pre/post hooks if an entity type is not "enabled" because they're all enabled, its just "enabling" unlocks  some additional functionality. That is that only if an entity type is enabled, you get Manage Fields, Manage Display, Manage Forms admin config pages, and drupal based view/add/edit/delete forms. 

Even if an entity type is not enabled, you get Views/Rules integration, and custom modules may use Drupal based hooks, which that and Rules need the pre/post hooks to function. So if we stop running these hooks when entity types not enabled, its a big upgrade issue for folks using more advanced configurations / custom functionality. 

The idea is to create new settings, one per entity type, and allow site admins to disable pre/post hooks per entity type should they choose. Default would be that pre/post hooks still fire unless site admins consciously disable. 

To disable by default would require a complicated check on module update, or perhaps a new major version where we document a backward compatibility change. 

This patch could work fine for you or others in the meantime, while we work up this idea.

jackrabbithanna:
I worked up an alternative PR: https://github.com/eileenmcnaughton/civicrm_entity/pull/514
This provides settings to disable pre/post hooks per entity type. 
There's a new field set in the "Advanced Settings" fieldset called "Disable pre/post hooks per entity type"
Check some or all of the entity types you want to disable pre/post hooks for.

jackrabbithanna:
Closing this in favor of #514

PR #513, opened 2025-01-14
Use custom InOperator for Autocomplete-Select.

puresyntax71:
Overview
Add support for "Autocomplete-Select" for InOperator. This is so `civicrm_entity_in_operator` would be used. If it isn't used, views wouldn't show any results.

To reproduce it, you can add a custom field with "Autocomplete-Select" as the "Field Type" and include it as a filter in the views.

Before
No Results when filtering with fields that are "Autocomplete-Select".

After
Has results.


jackrabbithanna:
Merging for inclusion in 4.0.0-beta6

PR #515, opened 2025-01-21
Add hooks.

puresyntax71:
Replace procedular hooks with OOP hooks.


jackrabbithanna:
Putting it in the can!

PR #518, opened 2025-01-29
only set path base field if entity type is enabled

jackrabbithanna:
Overview
only set path base field if entity type is enabled


jackrabbithanna:
Follow up to https://github.com/eileenmcnaughton/civicrm_entity/pull/517

PR #519, opened 2025-02-14
Uninstall page throwing an error

puresyntax71:
Overview
This error is thrown when going to the /admin/modules/uninstall page:

`Exception: Cannot rewind a generator that was already run in IteratorIterator->rewind() (line 656 of /var/www/html/web/core/lib/Drupal/Core/Extension/ModuleInstaller.php)`

Before
Error is thrown

After
No more errors.

Technical Details
Implementation of decorating a service is different now.

### DIFF
--- a/civicrm_entity.services.yml
+++ b/civicrm_entity.services.yml
@@ -20,14 +20,6 @@ services:
     tags:
       - { name: event_subscriber }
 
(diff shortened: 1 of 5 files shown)


jackrabbithanna:
I've manually tested this, and it fixes the error on the uninstall page

PR #525, opened 2025-03-13
CustomEntityField update constructor to remove deprecation warnings

jackrabbithanna:
Overview
https://www.drupal.org/project/civicrm_entity/issues/3512180

Before
----------------------------------------
Deprecation warning 

After
----------------------------------------
No warning

PR #527, opened 2025-03-24
Update DI to avoid changes in constructor.

puresyntax71:
This is in relation to #525 and https://www.drupal.org/project/civicrm_entity/issues/3512180 to avoid changes in the constructor.


jackrabbithanna:
Merging this in favor of #525 . Test fails unrelated. Did manual testing in D10 and D11

PR #533, opened 2025-08-20
Add a Distance field to Views for Proximity filters

dwoods-encircle:
Overview
We have a client who needs the distance showing for proximity searches from Civi Addresses.

https://www.drupal.org/project/civicrm_entity/issues/3541996#comment-16233331

Before
No Distance available for view results based on proximity filter origin.

After
A field called "Distance" is available, respecting the Units chosen in the filter.
A sort option called "Distance" is available, to order summary results in a table by distance.

Technical Details
See PR

Comments

Release notes snippet
The field and sort options are both called "Distance" available with CiviCRM_Address entity.


jackrabbithanna:
I tested this and it is functional as advertised. Thank you for your contribution!

dwoods-encircle:
An absolute pleasure! :)

PR #534, opened 2025-09-04
modify view permissions for countries and states-provinces

adixon:
From "view all contacts" to "access CiviCRM".

As noted in the issue: https://www.drupal.org/project/civicrm_entity/issues/3544960


jackrabbithanna:
We discussed on the drupal.org issue and this makes sense. The change will be noted in the release notes.

PR #537, opened 2025-09-09
Views argument default for contact ID; New option to limit cid to own contact for argument_validator/CivicrmContact

gngn23:
Overview
----------------------------------------
* New possibility to set views contextual filter for Contact ID
* Added validator for contextual filter to limit for currently logged in user`s Contact

Contains fix "use uf_id to match user-id in UFMatch" from #536

Before
----------------------------------------
See #536 for discussion of `uf_id` in Views argument default "Contact subtype from logged in user".

After
----------------------------------------
This proposal adds a new contextual filter to get data from a CiviCRM contact ID.

Technical Details
----------------------------------------
I added new `src/Plugin/views/argument_default/ContactId.php` and moved stuff from `ContactSubtype.php` to this new class:
* `$currentUser` and `$civicrmApi`
* `__construct() and `create()`
`ContactId::getArgument()` returns contact ID of current user (or 0 if none found).

With this I simplified `ContactSubtype`:
* extends `ContactId`
* thus inherits `$currentUser`, `$civicrmAp`i, `__construct()` and `create()`
* `ContactSubtype::getArgument()` now calls `parent::getArgument()` to get the cid (and then gets the subtype as before)

I also added a new option 'limit_own_contact' to `src/Plugin/views/argument_validator/CivicrmContact.php` which 
limits to current user\'s own contact ID. 

Comments
----------------------------------------
Included the change from #536

Release notes snippet
----------------------------------------
Enhanced views integration: new contextual filter Contact-ID, possibility to limit to currently logged in user`s contact.

PR #538, opened 2025-09-17
Views argument_default cid; limit cid to own contact for argument_val…

gngn23:
Overview
Another try (after fuddling with branches) replacing #537 
- New possibility to set views contextual filter for Contact ID
- Added validator for contextual filter to limit for currently logged in user`s Contact

Based on current 4.0.x

Before
We cannot
- set views argument default "Contact ID from logged in user".
- limit argument_validator/CivicrmContact not only to a contact subtype but also to "Contact ID from logged in user"

After
Both above are possible now.

Technical Details
I added new `src/Plugin/views/argument_default/ContactId.php` and moved stuff from `ContactSubtype.php` to this new class:
* `$currentUser` and `$civicrmApi`
* `__construct() and `create()`
`ContactId::getArgument()` returns contact ID of current user (or 0 if none found).

With this I simplified `ContactSubtype`:
* extends `ContactId`
* thus inherits `$currentUser`, `$civicrmAp`i, `__construct()` and `create()`
* `ContactSubtype::getArgument()` now calls `parent::getArgument()` to get the cid (and then gets the subtype as before)

I also added a new option 'limit_own_contact' to `src/Plugin/views/argument_validator/CivicrmContact.php` which 
limits to current user\'s own contact ID. 

Comments
Replaces #536

Release notes snippet
Enhanced views integration: new contextual filter Contact-ID, possibility to limit to currently logged in user`s contact.

### DIFF
--- a/src/Plugin/views/argument_default/ContactId.php
+++ b/src/Plugin/views/argument_default/ContactId.php
@@ -1,0 +1,106 @@
+<?php
+
+namespace Drupal\civicrm_entity\Plugin\views\argument_default;
+
+use Drupal\Core\Cache\Cache;
+use Drupal\Core\Cache\CacheableDependencyInterface;
+use Drupal\Core\StringTranslation\TranslatableMarkup;
+use Drupal\views\Attribute\ViewsArgumentDefault;
+use Drupal\views\Plugin\views\argument_default\ArgumentDefaultPluginBase;
+use Symfony\Component\DependencyInjection\ContainerInterface;
+use Drupal\Core\Session\AccountProxy;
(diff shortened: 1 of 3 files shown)


jackrabbithanna:
Seems well done! I tested the original contact type argument, which continues to function as expected.

I tested the new contact id argument, which also works as advertised. The new validation option for "CiviCRM Contact" also works, limiting to logged in user's contact. 

Very well done, thank you for contributing!!

jackrabbithanna:
This is released in 4.0.2 on d.o

PR #539, opened 2025-09-30
Feature add plugins, fixes to align in Inline Entity Form upgrade

paula2388:
Testing on 4.0.x branch, in Drupal 10

Add a entity reference field to a node type
Reference CiviCRM Activity, and all bundles (based on Activity Type)
Try to change field formatter to Inline Entity Form - Complex (Single not available, another problem?)
See ajax error in developer console
Look in log and find:
Drupal\Component\Plugin\Exception\InvalidPluginDefinitionException: The "civicrm_activity" entity type did not specify a inline_form handler. in Drupal\Core\Entity\EntityTypeManager->getHandler() (line 256 of /var/www/web/distro10.skvare.com/web/core/lib/Drupal/Core/Entity/EntityTypeManager.php).

Inline Entity Form version 2


jackrabbithanna:
Tested in D10 and D11

